PR #1, opened 2021-11-13
feat: Animation exports

LukeWasTakenn:


resource/client/anims.lua
```diff
@@ -0,0 +1,29 @@
...
+        Wait(wait)
+        if wait > 0 then ClearPedSecondaryTask(playerPed) end
+        RemoveAnimDict(dict)
+    end)
+end
+
+function PlayAnimAdvanced(wait, dict, name, posX, posY, posZ, rotX, rotY, rotZ, introSpeed, outroSpeed, duration, flag, time)
+    local playerPed = PlayerPedId()
+	RequestAnimDict(dict)
+	CreateThread(function()
```
LukeWasTakenn, line 19: I have no idea what the hell happened to this indentation lol

thelindat:
I've made a decision and won't be setting up a wrapper for these, since there's little to no value. They're _okay_ to save a bit of time in a resource but not worth the effort of importing or exporting.

PR #2, opened 2021-11-14
feat(natives/animations): add imports for animations

sondrix:
I have not really tested it, but theoretically it should work. Also I may be wrong for doing it this way.


sondrix:
Will it work like this? 

`local anims <const> = import 'natives/animations'`

PR #4, opened 2021-12-12
feat(resource/streaming): set timeout length and return requested item on success

DokaDoka:
This is a bit of a mess since I've had it sitting around for about a month and I don't know how to split it up nicely at this point.

It's basically 2 parts:

- adds the timeout parameter 
allows the number of ticks before timeout to be specified, timeout is avoided if `nil`
20 ticks was rarely enough for vehicle spawns

- returns the requested item on success
basically just to return the vehicle hash so I don't have to `joaat` it twice
added to the other functions for consistency

I haven't tested the functions aside from `requestAnimDict` and `requestModel` but everything should still function

### DIFF
--- a/resource/streaming/client.lua
+++ b/resource/streaming/client.lua
@@ -1,57 +1,62 @@
-local function hasLoaded(fn, name)
-	local timeout = 0
+local function hasLoaded(fn, lType, name, limit)
+	local timeout = limit
 	while not fn(name) do
 		Wait(0)
-		timeout += 1
-		if timeout > 20 then return false end
+		if timeout then
+			timeout -= 1
+			if timeout == 0 then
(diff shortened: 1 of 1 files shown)


thelindat:
Maybe something like this would be a bit nicer (I'll let you test it, don't want to load up FiveM atm).
```lua
local function hasLoaded(fn, type, request, timeout)
    timeout = timeout or 100
    while not fn(request) do
        Wait(0)
        timeout -= 0
        if timeout < 1 then
            return print(('Unable to load %s after %s ticks (%s)'):format(type, timeout, request))
        end
    end
    return name
end
```

DokaDoka:
100 ticks as the default is fine by me, will still need use a separate variable for timeout or it will always print 0

PR #5, opened 2021-12-23
fix(resource/streaming): use hasLoaded properly

DokaDoka:


thelindat:
pretty embarrassing that you committed such a failure

PR #7, opened 2022-02-10
feats(exports/entities): persistent tyres, doors, windows

sondrix:
I've tested persistent windows on every possible build and it seems there is something wrong with the window IDs so I left those that work for me.


sondrix:
Should I refactor the tyres a bit? I don't like the way I did it.

PR #10, opened 2022-03-25
GetVehicleXenonLightsColour -> GetVehicleXenonLightsColor

xrtw:
https://docs.fivem.net/natives/?_0x3DFF319A831E0CDB


thelindat:
Either should work (see old native names). It's fun when most natives seem to use _English_ but there are random ones using American English. Is it actually not working (pretty sure I've tested 🤔).

dolutattoo:
I can confirm that this is only working with `GetVehicleXenonLightsColor`
If you remember, I pull requested this change on your esx legacy fork

thelindat:
image
image

Both work fine. I'll merge if only for the native declarations, not that it matters.

PR #18, opened 2022-05-02
feat(web/dialog): Add password option to input

antond15:
Added password option to text input with toggleable show state.
image
image

### DIFF
--- a/web/src/features/dialog/InputDialog.tsx
+++ b/web/src/features/dialog/InputDialog.tsx
@@ -7,12 +7,15 @@ import {
   ModalBody,
   Box,
   Input,
+  InputGroup,
+  InputRightElement,
   Text,
   Button,
   Checkbox,
   Select,
 } from "@chakra-ui/react";
 import React from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { useNuiEvent } from "../../hooks/useNuiEvent";
 import { useLocales } from "../../providers/LocaleProvider";
 import { debugData } from "../../utils/debugData";
@@ -26,6 +29,7 @@ interface Props {
     type: RowType;
     label: string;
     options?: { value: string; label: string }[];
+    password?: boolean;
   }[];
 }
 
@@ -37,7 +41,7 @@ debugData<Props>([
       rows: [
         { type: "input", label: "Locker number" },
         { type: "checkbox", label: "Some checkbox" },
-        { type: "input", label: "Locker PIN" },
+        { type: "input", label: "Locker PIN", password: true },
         { type: "checkbox", label: "Some other checkbox" },
         {
           type: "select",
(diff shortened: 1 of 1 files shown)


web/src/features/dialog/InputDialog.tsx
```diff
@@ -104,9 +112,26 @@ const InputDialog: React.FC = () => {
...
-                    />
+                    <InputGroup>
+                      <Input
+                        onChange={(e) => handleChange(e.target.value, index)}
+                        type={!row.password || showPassword ? 'text' : 'password'}
+                      />
+                      {row.password && (
+                        <InputRightElement
+                          cursor="pointer"
+                          onClick={handleShowPassword}
```
LukeWasTakenn, line 123: Wouldn't this show the password on all input fields that have the password property on them?

web/src/features/dialog/InputDialog.tsx
```diff
@@ -104,9 +112,26 @@ const InputDialog: React.FC = () => {
...
+                          children={
+                            <FontAwesomeIcon
+                              fixedWidth
+                              icon={showPassword ? "eye" : "eye-slash"}
+                              fontSize="1em"
+                              style={{ paddingRight: 8 }}
+                            />
```
LukeWasTakenn, line 130: Don't really like the look of the FA icon on this, maybe just a button from Chakra UI's [example](https://chakra-ui.com/docs/components/form/input#password-input-example) would be enough?

LukeWasTakenn (changes requested):

web/src/features/dialog/InputDialog.tsx
antond15: Yees, fixed in the next commit.

web/src/features/dialog/InputDialog.tsx
antond15: Like this?
image
image

antond15: I think the icon is better, because every language has a different length of "Show" and "Hide" which would change the input text space a lot - even when switching between states.

LukeWasTakenn: > I think the icon is better, because every language has a different length of "Show" and "Hide" which would change the input text space a lot - even when switching between states.

Yeah you're right.

web/src/features/dialog/InputDialog.tsx
```diff
@@ -62,7 +66,15 @@ const InputDialog: React.FC = () => {
   const [visible, setVisible] = React.useState(false);
   const { locale } = useLocales();
 
+  const [passwordStates, setShowPassword] = React.useState<Array<boolean>>([]);
```
LukeWasTakenn, line 69: Use conventional naming for states please, `[variableName, setVariableName]`

LukeWasTakenn:
LGTM, thank you!

PR #32, opened 2022-05-25
Serbia Locales

DarkBoy621:


LukeWasTakenn:
Since it's so similar to existing Croatian, does it even need to be it's own locale?

DarkBoy621:
Ok np 🙂

PR #33, opened 2022-05-27
feat(circleprogress): Add Labels

SamShanks1:
Allows the option for text/label under the circular progress bar

### DIFF
--- a/web/src/features/progress/CircleProgressbar.tsx
+++ b/web/src/features/progress/CircleProgressbar.tsx
@@ -4,12 +4,14 @@ import {
   CircularProgressLabel,
   Flex,
   ScaleFade,
+  Text,
 } from "@chakra-ui/react";
 import { useNuiEvent } from "../../hooks/useNuiEvent";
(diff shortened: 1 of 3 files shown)


web/src/features/progress/CircleProgressbar.tsx
```diff
@@ -100,6 +105,18 @@ const CircleProgressbar: React.FC = () => {
...
           </CircularProgressLabel>
+          <Text
+              fontFamily="Inter"
+              isTruncated
+              fontSize={18}
+              fontWeight="light"
+              position="absolute"
+              top="115%"
+              left="50%"
+              transform="translate(-50%, -50%)"
```
LukeWasTakenn, line 116: Why not just use flexbox and avoid this weird aligning? 🤷

SamShanks1: Wrapped it in a flex and centred it

LukeWasTakenn (approved):

PR #56, opened 2022-07-09
Fixed fuel problem

Tinky124:


deputystrand:
is this the fix for random fuel level spawn when taking vehicle out from garage?

PR #57, opened 2022-07-12
feat(addCommand): Allow multiple groups

jag3dagster:
* Add ability to pass in a list of groups that can utilize the command being added

### DIFF
--- a/imports/addCommand/server.lua
+++ b/imports/addCommand/server.lua
@@ -39,7 +39,6 @@ end
 ---@param parameters table
 local function addCommand(group, name, callback, parameters, help)
 	if not group then group = 'builtin.everyone' end
-	if group and type(group) ~= 'table' then group = { group } end
 
 	if type(name) == 'table' then
 		for i = 1, #name do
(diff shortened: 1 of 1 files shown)


thelindat:
Would be better to check if groups is a table and then iterate over the values, else use the current behaviour.

jag3dagster:
Yeah, I was in a toss up over the two choices myself. I figured the overhead for the since element loop was relatively negligible, since it'll be repeated code in each block, but I'm okay with swapping to check instead of forcing it into a table.

thelindat:
> I figured the overhead for the since element loop was relatively negligible

It is, but I try to avoid short-lived objects and ipairs when I can. It does simplify things though. A minor improvement over the current checks on line 41/42.
```lua
if type(group) ~= 'table' then
    group = { group or 'builtin.everyone' }
end
```

jag3dagster:
> I try to avoid short-lived objects and ipairs when I can

Just out of curiosity, why do you avoid them? Just for cleanliness / readability? Or something else?

thelindat:
I prefer the look of `for i` compared to `ipairs`, it can be more flexible, and makes it obvious the table should be an array.
Once you get into large tables it also becomes an optimisation thing, and I'd prefer to be consistent with how I iterate tables.

I avoid short-lived tables to avoid extra memory allocation and gc cycles.

Both of these are negligible in this situation since the function only executes once, rather than frequently called in a loop or something; but again, I like to be consistent.

PR #59, opened 2022-07-27
fix: version verification

JustGodWork:
Prevent the script from crashing when version is not present in fxmanifest.lua

### DIFF
--- a/resource/version/server.lua
+++ b/resource/version/server.lua
@@ -8,7 +8,7 @@ function lib.versionCheck(repository)
 			response = json.decode(response)
 			if response.prerelease then return end
 
-			local currentVersion = GetResourceMetadata(resource, 'version', 0):match('%d%.%d+%.%d+')
+			local currentVersion = GetResourceMetadata(resource, 'version', 0) and GetResourceMetadata(resource, 'version', 0):match('%d%.%d+%.%d+')
 			if not currentVersion then return end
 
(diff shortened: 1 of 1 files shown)


thelindat:
Yes removing version is bigbrain

JustGodWork:
Checking currentVersion as you do was a good idea but if the version is not present and you try to use string.match an error appears, that's why I use a Ternary

JustGodWork:
Also you can use ```lua
if not currentVersion then return print("No version found in fxmanifest, please check you files or visit https://api.github.com/repos/overextended/ox_lib/releases/latest") end
```

thelindat:
> Checking currentVersion as you do was a good idea but if the version is not present and you try to use string.match an error appears, that's why I use a Ternary

If your script is calling lib.versionCheck then the assumption is you use version I fxmanifest and use GitHub releases to compare tags. If you’re removing version, remove the check.

thelindat:
> Also you can use
> 
> ```lua
> if not currentVersion then return print("No version found in fxmanifest, please check you files or visit https://api.github.com/repos/overextended/ox_lib/releases/latest") end
> ```

This message doesn’t make sense if we wanted a more sensible message in place of the current one. This function is exported for other resources to utilise, not just ox lib.

JustGodWork:
Maybe push your last code is great to. And yes your message is better mine was just for example 👍

PR #60, opened 2022-07-29
Feat/list menu

LukeWasTakenn:
WIP simplistic list menu

Current design (still subject to change):

image


thelindat:
## Bugs
- Debounce prevents onChange callbacks triggering on rapid changes
  - i.e. changing scrolling index from 1 -> 7, then moving to a different button only triggers a change for the new button
- Sending an empty table in options renders an empty box
- Sending a string in options renders an empty box, which cannot be selected
- Causing an error with invalid input breaks the component (need to restart ox_lib)
- Window does not scroll at set increments (i.e. move down one button)

## Enhancements
- Return arguments
  - QoL, recommend to register menu and store in the resource; then you can reference any values with index
- Scrollbar or icon to show more options are available
- Set default scroll index
  - sample use case: building a vehicle-modding menu, need current mods to be set as default
  - maybe a way to pass this through without registering the menu again?
- Edit menus
  - like mentioned above, a way to pass changes without registering the complete menu?
- where close button

ultrahacx:
To
Mr. Luke
    Please add a close button.
Regards,
Linden

PR #66, opened 2022-08-04
feat(web/dialog): Support default values

antond15:
- Placeholders for input and number type
- Default values for input, number and checkbox type

image


LukeWasTakenn (approved):
Works well, do you want to do default for select as well?

You'd have to pass one of the values to a `default` property then iterate the options array and if it matches one of the values call the handleChange function for that value, otherwise just don't set any defaults.

Could you also please update the documentation?

antond15:
> Could you also please update the documentation?

👍 https://github.com/overextended/overextended.github.io/pull/35

LukeWasTakenn (approved):
Awesome, thank you so much.

PR #67, opened 2022-08-08
fix(web/dialog): Setting non-selectable placeholder

antond15:
React doesn't like the `selected` attribute (causing a warning in web console - not in-game).

`defaultValue` should be used instead so there is no need to create placeholder option if `defaultValue` is already defined by row options.

### DIFF
--- a/web/src/features/dialog/components/select.tsx
+++ b/web/src/features/dialog/components/select.tsx
@@ -23,16 +23,16 @@ const SelectField: React.FC<Props> = (props) => {
     <>
       <Box mb={3} key={`select-${props.index}`}>
         <Select
-          onChange={(e) => props.handleChange(e.target.value, props.index)}
-          defaultValue={props.row.default}
+          onChange={(e: any) => props.handleChange(e.target.value, props.index)}
+          defaultValue={props.row.default || ""}
         >
           {/* Hacky workaround for selectable placeholder issue */}
-          <option value="" selected hidden disabled>
+          {!props.row.default && (<option value="" hidden disabled>
             {props.row.label}
-          </option>
+          </option>)}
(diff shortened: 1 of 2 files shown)


web/src/features/dialog/components/select.tsx
```diff
@@ -23,16 +23,16 @@ const SelectField: React.FC<Props> = (props) => {
     <>
       <Box mb={3} key={`select-${props.index}`}>
         <Select
-          onChange={(e) => props.handleChange(e.target.value, props.index)}
-          defaultValue={props.row.default}
+          onChange={(e: any) => props.handleChange(e.target.value, props.index)}
```
LukeWasTakenn, line 26: Why explicitly type this as `any`?

The type was already getting inferred, don't see a reason to explicitly type it.

LukeWasTakenn:
You might want to rebase the branch to not commit linden's commit that was force pushed away.

web/src/features/dialog/components/select.tsx
antond15: Should I use `React.ChangeEvent<HTMLSelectElement>` instead or just keep it implicitly as `any`?

LukeWasTakenn: > Should I use `React.ChangeEvent<HTMLSelectElement>` instead or just keep it implicitly as `any`?

You should be able to just remove the type completely and it should get inherited.

Otherwise you can use `React.ChangeEvent<HTMLSelectElement>`.

antond15: Without any explicit type, it is `any`, so I used the React one.

LukeWasTakenn (approved):
Tested, lgtm, thank you!

PR #68, opened 2022-08-08
refactor(web/dialog): Split interfaces

antond15:
For better typing of row options for different input types.
Eg. various components accepts different value types for `defaultValue`, so it should be split for each input type.


LukeWasTakenn (approved):
While I really dislike the `I` prefix for interfaces it probably makes sense here since I'm really not sure how I'd name them otherwise.

Thank you again.

PR #69, opened 2022-08-10
feat(web/dialog): Add slider input

antond15:
Slider example


LukeWasTakenn (approved):
Works great as usual,

would you mind updating the docs as well please?

PR #70, opened 2022-08-12
feat(progress): Add interrupt option

jag3dagster:
Add the ability to specify if an action can be interrupted. Options that I've thought of off the bat are ped going ragdoll or ped being handcuffed.


resource/interface/client/progress.lua
```diff
@@ -40,32 +40,48 @@ local function startProgress(data)
...
+					if IsPedRagdoll(cache.ped) then
+						progress = false
+					end
+				end
+
+				if data.interrupt.cuff then
+					if IsPedCuffed(cache.ped) then
+						progress = false
+					end
+				end
```
thelindat, line 84:
```lua
if data.interrupt.ragdoll and IsPedRagdoll(cache.ped) then
    progress = false
end

if data.interrupt.cuff and IsPedCuffed(cache.ped) then
    progress = false
end
```

As above would be better. These would likely also be better off as defaults (so you have to explicitly set them to false to disable), and should get checked before the progress bar begins.

thelindat (changes requested):

resource/interface/client/progress.lua
jag3dagster: Updated based on feedback

jag3dagster:
Ope, realized I overlooked something... You can hold off on the review for a sec

jag3dagster:
Alright, fixed what I overlooked and added falling to potential interrupts to ignore.

Interesting thing I realized during testing of this though was that ox_inventory useItem is already checking for dead / ragdoll / falling before attempting to use an item. Would y'all want that updated to be a check for ignoreInterrupt in client table defaulting to current behavior if no ignore is found?

PR #71, opened 2022-08-16
feat(imports): getClosestVehicle and getNearbyVehicles

jag3dagster:
Definitely would find these useful, but am curious if it would be worthwhile to find a more generic way to handle getters such as these and the player equivalents.  It may be too much of a headache to be worth the generalization of the functions though.


thelindat:
That would require formatting as a module rather than returning a single function, i.e.
```lua
local getClosest = {}

function getClosest.Player(...) end
function getClosest.Vehicle(...) end

return getClosest
```

Or otherwise importing common code into each import (and referencing an already loaded version). Basically the way the standard Lua_require works, or require and import in JS.

---

I prefer loading single functions when possible, so the second is preferred; the benefit being a slight reduction in code repetition. Overall it's very minor and not something I'm interested in currently.

PR #75, opened 2022-08-17
tweak(web/list): colour scheme

Skryptific:
Tweaked the colour scheme of the list menu to match the context menu.

Screenshot_2


LukeWasTakenn:
Instead of merging this now, later on down the line I'll be adding theme support to the lib.

Idea is to have dark, dark alternative (which is this blue/gray color scheme) and light themes and to store the choice in local storage so that it can be accessed from other nui resources as well, though I haven't played around with local storage at all yet so not sure if that would even work.

Issue #77, opened 2022-08-25
inputDialog on menu callback


Aaron-Downham:
In the callback function on the lib menu, I'm trying to request an input from the player using lib.inputDialog however the input only appears for a split second before closing. Using v2.10.3 and if I print the input it's just an empty table; below is an example bit of code, just select the option in the menu and you should be able to reproduce the issue.

```
lib.registerMenu({
    id = "testMenu",
    title = "Test Menu",
    options = {
        {label = "Test"}
    }
}, function()
    local input = lib.inputDialog("Test", {"Input"})
    print(json.encode(input))
end)

RegisterCommand("test", function()
    lib.showMenu("testMenu")
end)
```

LukeWasTakenn:
Seems to be happening because the input dialog is being set to visible and the confirm is being triggered from the enter key.

https://github.com/overextended/ox_lib/blob/4bd0e6936def637a80a0857fb603ea64cafec705/web/src/features/dialog/InputDialog.tsx#L67-L69

LukeWasTakenn:
Should be resolved in commit ae25d109fc5127d8c84a91f98edd0c67382eeff2

PR #91, opened 2022-09-10
fix(resource/input): dialog not closing on first confirm

BerkieBb:
Moving the promise of `input` to another variable and making it nil returned the right value to lib.inputDialog but it didn't let the dialog close on confirm for some reason, once the promise is resolved directly on `input`, the UI does close on confirm.


LukeWasTakenn:
The changes to this were made for #77

Worked fine when I was testing it 🤷

BerkieBb:
I'm actually dumb lol, I made a recursion by accident on lib.inputDialog once it's value is returned 🤦‍♂️

PR #98, opened 2022-09-16
feat(menu): Menus update live if the openMenu is updated

jag3dagster:
* Update the menu live if `lib.setMenuOptions` is called for the current `openMenu`


thelindat:
lib.setMenuOptions isn't really intended for super-reactive menus you can constantly update while going through.
Just make the changes and re-open the menu if that's necessary; if you can't already send the position to open the menu, add that.

jag3dagster:
I had a feeling that'd be the feedback. Will work on a default list index for open menu in a different PR.

jag3dagster:
New PR

https://github.com/overextended/ox_lib/pull/99

PR #103, opened 2022-09-20
feat(imports/points): nearby distance override

BerkieBb:
Adds an override for the distance to trigger point:nearby()

### DIFF
--- a/imports/points/client.lua
+++ b/imports/points/client.lua
@@ -1,6 +1,7 @@
 ---@class CPoint
 ---@field id number
 ---@field coords vector3
+---@field nearbyDistance number
 ---@field distance number
 ---@field currentDistance number
 ---@field remove fun()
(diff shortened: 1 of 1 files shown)


thelindat:
This is what point.distance is used for, it's just not very descriptive and I can't go back and change it (compatibility).

point.distance = range to trigger point:nearby (which you can then use for DrawMarker, IsControlPressed, etc)
point.currentDistance = how far you are from the point, calculated when you're already in range

Unless I'm a potato

BerkieBb:
I use this to trigger nearby earlier, so that way I can draw a marker first and then make my own loop when onEnter is triggered, easy alternative compared to needing to check point.currentDistance to draw when you're nearby instead of a function already doing that

BerkieBb:
Now point.distance is used for both onEnter and nearby, so basically nearby acts as inside actually

thelindat:
You would just DrawMarker without checking `self.distance`, then use `self.distance` for control checks or such.
image

Making another loop when onEnter triggers is a bit redundant since that's what nearby is (and nearby is just "inside").
- onEnter: trigger when you get in range
- nearby: trigger each tick while in range

BerkieBb:
Yup, you're right, that's much better, thanks!

thelindat:
> You would just DrawMarker without checking `self.currentDistance`, then use `self.currentDistance` for control checks or such. image

I can probably add `self.isClosest` instead of needing to compare `self.id` to lib.points.closest().id` as well, as it's a bit verbose and confusing.

If you have anything you'd like improved on points let me know but they're pretty much just a generic and super-optimised distance check which work very well for markers. lib.zones.sphere actually functions pretty much identically (with some overhead).

BerkieBb:
self.isClosest would be great yeah

PR #108, opened 2022-09-26
Check whether zone.inside function exists

tony-stark-17:


thelindat:
If zone.debug and zone.inside aren't defined then the zone shouldn't even be inside this loop.

thelindat:
Any follow up, i.e. I missed something; or should I close this? @tony-stark-17

PR #110, opened 2022-09-29
Revert "fix(version): properly compare minor/patch semver"

D4isDAVID:
The following version check fails:
- minimum version `2.11.2`
- current version `2.11.1`

Thus, the following version check should also fail:
- minimum version `2.11.10`
- current version `2.11.1`

But with the current way of checking the version, this actually passes.
I didn't directly check this through a script, but this bug should happen since `0.1 == 0.10`.

Revert commit bdb88e74c5e80ce5a9bd6eea268f11970c6d4962.

### DIFF
--- a/resource/version/server.lua
+++ b/resource/version/server.lua
@@ -19,11 +19,13 @@ function lib.versionCheck(repository)
 			local latestVersion = response.tag_name:match('%d%.%d+%.%d+')
 			if not latestVersion or latestVersion == currentVersion then return end
 
-			local cMajor, cMinor = string.strsplit('.', currentVersion, 2)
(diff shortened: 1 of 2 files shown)


thelindat:
The previous version doesn't work either, so reverting doesn't resolve anything; and I would say it was worse.

thelindat:
https://github.com/overextended/ox_lib/commit/e624aa255996a658ba3a80496bdda1e818a6ffde

This should actually work.

D4isDAVID:
> [e624aa2](https://github.com/overextended/ox_lib/commit/e624aa255996a658ba3a80496bdda1e818a6ffde)
> 
> This should actually work.

Looks good to me.

PR #111, opened 2022-09-29
fix(web/features/menu/list): force close menu when using the nui event

BerkieBb:
This fixes the issue of canClose stopping the menu from being closed when using lib.hideMenu

### DIFF
--- a/web/src/features/menu/list/index.tsx
+++ b/web/src/features/menu/list/index.tsx
@@ -37,8 +37,8 @@ const ListMenu: React.FC = () => {
   const [indexStates, setIndexStates] = useState<Record<number, number>>({});
   const listRefs = useRef<Array<HTMLDivElement | null>>([]);
 
-  const closeMenu = (ignoreFetch?: boolean, keyPressed?: string) => {
-    if (menu.canClose === false) return;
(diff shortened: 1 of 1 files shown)


web/src/features/menu/list/index.tsx
```diff
@@ -37,8 +37,8 @@ const ListMenu: React.FC = () => {
   const [indexStates, setIndexStates] = useState<Record<number, number>>({});
   const listRefs = useRef<Array<HTMLDivElement | null>>([]);
 
-  const closeMenu = (ignoreFetch?: boolean, keyPressed?: string) => {
-    if (menu.canClose === false) return;
+  const closeMenu = (ignoreFetch?: boolean, keyPressed?: string, forceClose?: boolean) => {
```
LukeWasTakenn, line 40: Since forceClose isn't a param in Lua (and isn't really needed anyway) you can't you make this the default behavior and not have to add the extra param?

BerkieBb: Well, if it is default behaviour, then the canClose variable will be useless as it will always force it then when it's not supposed to do it on the usual escape/backspace press

LukeWasTakenn: Yeah true, good point.

PR #117, opened 2022-10-06
Add support for explicit nils in callbacks

Mkeefeus:
Lua tables do not support explicit nils. This causes problems if you have a callback that may want to return multiple values, one of which could be nil, like `return '1', nil, '3'` which after being packed and unpacked into a table, would return `'1' nil, nil` . I found this working on my own callback system before finding ox_lib, and found a solution here https://riptutorial.com/lua/example/15644/advanced-usage using a tuple function, which I have added to the ox_lib callbacks here.

### DIFF
--- a/imports/callback/client.lua
+++ b/imports/callback/client.lua
@@ -23,6 +23,19 @@ local function eventTimer(event, delay)
 	return true
 end
 
+---@param ... any
+---@return function
+local function tuple(...)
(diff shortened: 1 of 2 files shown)


Mkeefeus:
Found some errors I need to work out first

thelindat:
Coroutine tuple is an.. interesting solution; but handled better by using msgpack on the response.

PR #126, opened 2022-10-14
open state getter for textUI

BerkieBb:
Since the text ui can be overwritten using lib.showTextUI or hidden when it shouldn't, we should be able to check whether the textui is already open or not to prevent it


BerkieBb:
https://github.com/overextended/overextended.github.io/pull/56

PR #129, opened 2022-10-15
Skill check

LukeWasTakenn:


resource/interface/client/skillcheck.lua
```diff
@@ -0,0 +1,62 @@
...
+                gamePromise = nil
+                break
+            end
+        end
+    else
+        SendNUIMessage({
+            action = 'startSkillCheck',
+            data = difficulty
+        })
+    end
```
thelindat, line 48: Would be best to handle this in the ui; 'startSkillCheck' should await all promises to resolve before triggering 'skillCheckOver'.

thelindat (changes requested):

resource/interface/client/skillcheck.lua
thelindat: Also just noticed difficulty can be a string, map, or array (containing strings or maps), so sending a custom difficulty wouldn't work 🤔
```lua
lib.skillCheck({areaSize = 250, speedMultiplier = 2})
```
This would fix it, but should still be done in nui anyway.
```lua
if type(difficulty) == 'table' and table.type(difficulty) == 'array' then
```

LukeWasTakenn: It will all be solved once I move the logic into UI

PR #137, opened 2022-10-24
feat: vehicle class

Andyyy7666:
I need this for when getting the vehicle properties


thelindat:
why

thelindat:
That's not the point of the function..
Just get the vehicle class if you need that.

PR #138, opened 2022-10-24
tweak(vehicleProperties): add vehicle type

davidxd33:
Returns the vehicle type as part of the properties table.


davidxd33:
I'm an idiot

PR #142, opened 2022-10-26
feat(interface/dialog): readonly, disabled flags

davidxd33:
This adds support for making fields disabled or read only.

### DIFF
--- a/package/client/resource/interface/input.ts
+++ b/package/client/resource/interface/input.ts
@@ -10,7 +10,6 @@ interface InputDialogRowProps {
   placeholder?: string;
   default?: string | number;
   disabled?: boolean;
-  readonly?: boolean;
   checked?: boolean;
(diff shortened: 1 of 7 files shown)


LukeWasTakenn:
Are readonly and disabled not kind of the same thing?

davidxd33:
> Are readonly and disabled not kind of the same thing?

One greys it out and changes the cursor, the other just makes it "uninteractable" while looking normal.

LukeWasTakenn (changes requested):
The TS types in the package should also be updated, and could you also make a PR to the docs?

davidxd33:
> Member

Done

LukeWasTakenn (approved):
Removed read only property because I can already see people complaining that it's not working even though it's the intended behavior.

PR #146, opened 2022-10-29
feat(menu): checkbox

BerkieBb:
It works, but the style is not good, so I'll let frontend-god Luke finish this

### DIFF
--- a/web/src/features/menu/list/index.tsx
+++ b/web/src/features/menu/list/index.tsx
@@ -61,43 +61,28 @@ const ListMenu: React.FC = () => {
         });
         break;
       case 'ArrowRight':
-        if (Array.isArray(menu.items[selected].values)) {
+        if (Array.isArray(menu.items[selected].values))
           setIndexStates({
             ...indexStates,
             [selected]:
-              indexStates[selected] + 1 <= menu.items[selected].values?.length! - 1
-                ? indexStates[selected] + 1
-                : 0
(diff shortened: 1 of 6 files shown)


web/src/features/menu/list/index.tsx
```diff
@@ -59,41 +61,71 @@ const ListMenu: React.FC = () => {
...
+              indexStates[selected] - 1 >= 0
+                ? indexStates[selected] - 1
+                : menu.items[selected].values?.length! - 1
+          });
+        } else if (menu.items[selected].checked !== undefined) {
+          setCheckedStates({
+            ...checkedStates,
+            [selected]: !checkedStates[selected]
+          });
+        }
```
LukeWasTakenn, line 93: I don't know how I feel about left-right arrows checking/unchecking the checkbox, imo this should be handled by `Enter`, then just make it not close the menu in the process.

While left-right arrows to check/uncheck are cool it adds unnecessarily complexity to the code and most people's first instinct would be to press enter to toggle the checkbox.

BerkieBb: Well, both enter and left/right handle it

LukeWasTakenn: yes but not returning out of the confirm function doesn't give the user that feedback that the checkbox changed

BerkieBb: We've got the onCheck function for that

PR #151, opened 2022-11-05
feat(resource/cache): current weapon

BerkieBb:
may as well

### DIFF
--- a/resource/cache/client.lua
+++ b/resource/cache/client.lua
@@ -40,7 +40,7 @@ CreateThread(function()
 		local hasWeapon, currentWeapon = GetCurrentPedWeapon(ped, true)
 
 		if hasWeapon then
-			if not cache.weapon or cache.weapon ~= currentWeapon then
+			if cache.weapon ~= currentWeapon then
 				cache:set('weapon', currentWeapon)
 			end
 		else
--- a/resource/cache/client.lua
+++ b/resource/cache/client.lua
@@ -37,6 +37,16 @@ CreateThread(function()
 			cache:set('seat', false)
 		end
 
+		local hasWeapon, currentWeapon = GetCurrentPedWeapon(ped, true)
+
+		if hasWeapon then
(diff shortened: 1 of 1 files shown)


resource/cache/client.lua
```diff
@@ -37,6 +37,16 @@ CreateThread(function()
...
 		end
 
+		local hasWeapon, currentWeapon = GetCurrentPedWeapon(ped, true)
+
+		if hasWeapon then
+			if not cache.weapon or cache.weapon ~= currentWeapon then
+				cache:set('weapon', currentWeapon)
+			end
+		else
+			cache:set('weapon', false)
```
thelindat, line 47: Don't you technically always have a weapon (i.e. weapon_unarmed)?

resource/cache/client.lua
```diff
@@ -37,6 +37,16 @@ CreateThread(function()
 			cache:set('seat', false)
 		end
 
+		local hasWeapon, currentWeapon = GetCurrentPedWeapon(ped, true)
+
+		if hasWeapon then
+			if not cache.weapon or cache.weapon ~= currentWeapon then
```
thelindat, line 43: `cache.weapon ~= currentWeapon` should be enough 🤷

thelindat (changes requested):

resource/cache/client.lua
BerkieBb: If your currentWeapon is weapon_unarmed, hasWeapon will be false

BerkieBb: So actually reading that again, the if statement is probably not even needed at all

BerkieBb:
Accidentally stopped thinking

BerkieBb:
Nevermind, that commit was actually good as I just now saw that cache:set already checks for the previous value, so here comes the double revert

LukeWasTakenn (approved):

PR #152, opened 2022-11-08
Update fxmanifest.lua to rdr3 game

alphatule:


thelindat:
I know there's some stuff that won't work with RDR, but don't have the game to test.
Would be nice to have a list of unsupported features, or note in the modules themselves; or potentially workarounds so they work (i.e. lib.addKeybind).

PR #153, opened 2022-11-10
feat(resource/cache): previous value for onCache

BerkieBb:


thelindat:
Previous value will already be available in the resources using cache, simply compare the new value to `cache.ped` or such.
To otherwise make it more intuitive, add lastValue but make it part of the cache event handler; i.e.
```lua
local function triggerCacheHandler(key, value, cb)
    cb(value, cache[key])
end

lib.onCache = function(key, cb)
    AddEventHandler(('ox_lib:cache:%s'):format(key), cb)
end
```

PR #158, opened 2022-11-22
Add cached coords + Menu list design clean up

Scullyy:
Cached coords and cleaned up the menu list a little, this makes the menu list look less "bulky" and also fixes an issue with it overlapping with the fivem watermark when placed in the top right.

### DIFF
--- a/resource/cache/client.lua
+++ b/resource/cache/client.lua
@@ -10,6 +10,7 @@ function cache:set(key, value)
 	end
 end
 
+local GetEntityCoords = GetEntityCoords
 local GetVehiclePedIsIn = GetVehiclePedIsIn
 local GetPedInVehicleSeat = GetPedInVehicleSeat
 local GetVehicleMaxNumberOfPassengers = GetVehicleMaxNumberOfPassengers
@@ -18,7 +19,8 @@ local GetCurrentPedWeapon = GetCurrentPedWeapon
 CreateThread(function()
 	while true do
 		local ped = PlayerPedId()
-		cache:set('ped', ped)
+		local coords = GetEntityCoords(ped)
+		cache:set('coords', coords)
 
 		local vehicle = GetVehiclePedIsIn(ped, false)
 
--- a/resource/cache/client.lua
+++ b/resource/cache/client.lua
@@ -19,6 +19,7 @@ local GetCurrentPedWeapon = GetCurrentPedWeapon
 CreateThread(function()
 	while true do
 		local ped = PlayerPedId()
+		cache:set('ped', ped)
 		local coords = GetEntityCoords(ped)
 		cache:set('coords', coords)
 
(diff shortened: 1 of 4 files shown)


thelindat:
You should show the changes you're trying to make, I don't want to have to pull just to compare. Furthermore, mixing completely unrelated changes as a single commit _and PR_ is just bad practice and adds more work for everybody.

Cache isn't supposed to be used for values that change frequently, syncing coords every 100ms is undesirable. It _is_ however available when either zones or points have been loaded into a resource (just the variable, not the event handler; which would again be undesirable). You've also removed cache.ped entirely.

UI changes are just personal preference so not a "clean up", and having the top-left margin so much larger than other positions is weird af; the FiveM watermark can also appear in any corner, not _just_ the top-left.

Scullyy:
> You should show the changes you're trying to make, I don't want to have to pull just to compare. Furthermore, mixing completely unrelated changes as a single commit _and PR_ is just bad practice and adds more work for everybody.

Here is an image of the changes https://i.gyazo.com/1a135dedec591100dc3537e83dffebe6.png, just reduces the size of those big bulky boxes, adds the arrow to the description tooltip and removes the forced capitalization on the title as if people want that capitalized they can just type it that way.

> Cache isn't supposed to be used for values that change frequently, syncing coords every 100ms is undesirable. It _is_ however available when either zones or points have been loaded into a resource (just the variable, not the event handler; which would again be undesirable). You've also removed cache.ped entirely.

It has worked well for me the last few weeks I have been using it and I'm sure it'll work well for others, when I ran benchmarks when I added it awhile back it did improve the performance on my resources but that's obvious due to it being cached.

> UI changes are just personal preference so not a "clean up", and having the top-left margin so much larger than other positions is weird af; the FiveM watermark can also appear in any corner, not _just_ the top-left.

Yes but from reading comments people have been making regarding the menu this is a personal preference of the majority, in both mine and many others opinion the menu options are just very bulky and do not need to be that big. Also the fivem watermark doesn't go into any corner, it is limited to the right side of the screen now if you look at cfx's fivem repository. I've been using it like this for weeks and it doesn't look weird at all.

LukeWasTakenn:
> Here is an image of the changes https://i.gyazo.com/1a135dedec591100dc3537e83dffebe6.png,

Personally I prefer the current design a lot more.

Scullyy:
> Personally I prefer the current design a lot more.

It essentially looks the same just less bulky which I’ve seen a lot of people complain about. This also doesn’t force capitalisation which I’m not sure why it ever did as if people wanted them they can just type it that way.

LukeWasTakenn:
> which I’m not sure why it ever did

Design choice, but yeah I don't mind not forcing it.

> looks the same just less bulky

Which is exactly why I prefer the current one haha

thelindat:
Really just need to move styles to css so people can easily edit without having to build the entire ui.

Scullyy:
Yeah I think removing the forced capitalized titles and moving everything to css would be better

PR #162, opened 2022-11-24
Add RedM support

Scullyy:


resource/cache/client.lua
```diff
@@ -38,6 +40,13 @@ CreateThread(function()
 			cache:set('seat', false)
 		end
 
+		if cache.game == 'redm' then
+			local mount = GetMount(ped)
+			local onMount = IsPedOnMount(ped)
+			cache:set('mounted', onMount)
+			cache:set('mount', onMount and mount or false)
+		end
```
thelindat, line 48: Probably my only complaint; for vehicle the value is a number or false when not in a vehicle. No need for "mounted".

thelindat (changes requested):

resource/cache/client.lua
Scullyy: Is the mount native the same way? Mounts aren’t considered vehicles. The vehicles are the wagons.

Scullyy: Was still waking up when I first responded to this but see what you mean, the mounted is pointless because mount is false if the player isn't on one so I removed it

PR #166, opened 2022-12-03
[Menu] Progress Bars

ItsANoBrainer:
Adds support for progress bars to items in both context and list menus. 

-  `progress` key in the option. It will be a number between 0 and 100 (inclusive) as a percentage.
-  `colorScheme` key in the option for the progress bar. Defaults to gray.
- `iconColor` added to menu

image
image

### DIFF
--- a/web/src/features/menu/context/Item.tsx
+++ b/web/src/features/menu/context/Item.tsx
@@ -10,7 +10,7 @@ import {
   Spacer,
   Image,
   HStack,
-  Progress
+  Progress,
 } from '@chakra-ui/react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Option, ContextMenuProps } from '../../../interfaces/context';
@@ -73,13 +73,19 @@ const Item: React.FC<{
                     {button.title ? button.title : buttonKey}
                   </Text>
                 </Box>
-                { button.description && (
-                  <Box paddingBottom={1} color={button.disabled ? '#718096' : undefined}">
+                {button.description && (
+                  <Box paddingBottom={1} color={button.disabled ? '#718096' : undefined}>
                     <Text>{button.description}</Text>
                   </Box>
                 )}
-                { button?.progress && (
-                    <Progress value={button.progress} size="sm" colorScheme={button.colorScheme || "gray"} borderRadius="md" marginRight="5px"/>
+                {button?.progress && (
+                  <Progress
+                    value={button.progress}
+                    size="sm"
+                    colorScheme={button.colorScheme || 'gray'}
+                    borderRadius="md"
+                    marginRight="5px"
+                  />
                 )}
               </Box>
               {(button.menu || button.arrow) && button.arrow !== false && (
--- a/web/src/features/menu/context/Item.tsx
+++ b/web/src/features/menu/context/Item.tsx
@@ -79,7 +79,7 @@ const Item: React.FC<{
                   </Box>
                 )}
                 { button?.progress && (
-                    <Progress value={button.progress} size="sm" colorScheme={button.colorScheme ? button.colorScheme : "gray"} borderRadius="md" marginRight="5px"/>
+                    <Progress value={button.progress} size="sm" colorScheme={button.colorScheme || "gray"} borderRadius="md" marginRight="5px"/>
                 )}
               </Box>
               {(button.menu || button.arrow) && button.arrow !== false && (
--- a/web/src/features/menu/context/Item.tsx
+++ b/web/src/features/menu/context/Item.tsx
@@ -10,6 +10,7 @@ import {
   Spacer,
   Image,
   HStack,
+  Progress
 } from '@chakra-ui/react';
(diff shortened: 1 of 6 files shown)


web/src/features/menu/context/Item.tsx
```diff
@@ -66,13 +67,15 @@ const Item: React.FC<{
...
                 />
               )}
-              <Box>
+              <Box w="100%">
                 <Box paddingBottom={button.description ? 1 : 0}>
                   <Text w="100%" fontWeight="medium" color={button.disabled ? '#718096' : undefined}>
                     {button.title ? button.title : buttonKey}
                   </Text>
                 </Box>
-                {button.description && (
```
LukeWasTakenn, line 75: Why not make it so you can have a progress bar *and* a description as well? 🤷

web/src/features/menu/list/ListItem.tsx
```diff
@@ -60,6 +60,11 @@ const ListItem = forwardRef<Array<HTMLDivElement | null>, Props>(({ item, index,
             <Text>{item.label}</Text>
             <CustomCheckbox checked={checked}></CustomCheckbox>
           </Flex>
+        ) : item.progress !== undefined ? (
+          <Flex flexDirection="column" w="100%" marginRight="5px">
+            <Text verticalAlign="middle" marginBottom="3px"> {item.label} </Text>
+            <Progress value={item.progress} size="sm" colorScheme={item.colorScheme} borderRadius="md"/>
+          </Flex>
```
LukeWasTakenn, line 67: Maybe instead of adding the progress bar at the bottom and normal text at the top you could try how it's like for side scrolling buttons, where you have smaller faint text at the top and then progress bar at the bottom.

Could look nicer but it may leave some awkward empty space in between 🤔

web/src/features/menu/list/index.tsx
```diff
@@ -11,6 +11,8 @@ import React from 'react';
 
 export interface MenuItem {
   label: string;
+  progress?: number;
+  colorScheme?: string;
```
LukeWasTakenn, line 15: The list menu is kinda built to be colourless, but since I'll be moving all the styling to scss eventually I wouldn't mind letting people pick the colour.

Though it should definitely have a default that matches the current colours of the menu.

LukeWasTakenn (changes requested):
Overall great looking PR, though I am worried people wouldn't know they can press the progress bar buttons 🤔

web/src/features/menu/list/index.tsx
ItsANoBrainer: Added default of gray, matches the UI well. Forgot about having this default.

Regarding colors, I was thinking about adding a key for coloring each item individually, or just the progress bar and icon on their own, but didn't know if this would start bloating up the options.

Thoughts on color scheming system?

web/src/features/menu/list/ListItem.tsx
ItsANoBrainer: Doesnt look too bad, but for my specific use case I like having it with a heavier font weight.

image

web/src/features/menu/context/Item.tsx
ItsANoBrainer: I thought it would bloat too much, but adjusted it to support if people want it!

web/src/features/menu/list/ListItem.tsx
LukeWasTakenn: Bolder definitely looks better

web/src/features/menu/list/index.tsx
LukeWasTakenn: > Thoughts on color scheming system?

I think `colorScheme` for progress bars is fine, *maybe* even `iconColor` but I wouldn't want people to be able to colour the whole buttons as I could see some horrible creations manifest out of that 😞

web/src/features/menu/context/Item.tsx
```diff
@@ -66,17 +67,20 @@ const Item: React.FC<{
...
+              <Box w="100%">
+                <Box>
                   <Text w="100%" fontWeight="medium" color={button.disabled ? '#718096' : undefined}>
                     {button.title ? button.title : buttonKey}
                   </Text>
                 </Box>
-                {button.description && (
-                  <Box paddingBottom={1} color={button.disabled ? '#718096' : undefined}>
+                { button.description && (
+                  <Box paddingBottom={1} color={button.disabled ? '#718096' : undefined} fontWeight="light">
```
LukeWasTakenn, line 77: Not a fan of the lighter look on the font at all, again, this is something people will be able to change more easily when everything is moved over to scss 🤷

LukeWasTakenn (changes requested):
Should be pretty much good to go after this I think

web/src/features/menu/list/index.tsx
ItsANoBrainer: True, I agree, could get a bit crazy.

LukeWasTakenn (approved):
Thanks for the PR 👍 

Could you also please update the docs?

PR #172, opened 2022-12-13
fix(web/menu): Combine initial select and first render checks

jag3dagster:
* Combining these checks should fix an issue of having to toggle a checkbox multiple times before it is recognized as being toggled

### DIFF
--- a/web/src/features/menu/list/index.tsx
+++ b/web/src/features/menu/list/index.tsx
@@ -50,6 +50,7 @@ const ListMenu: React.FC = () => {
   };
 
   const moveMenu = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    if (firstRenderRef.current) firstRenderRef.current = false
     switch (e.code) {
       case 'ArrowDown':
         setSelected((selected) => {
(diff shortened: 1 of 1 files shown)


LukeWasTakenn (changes requested):
Seems to be triggering onSideScroll on open if the first item in the menu is a side scroll button 😭

jag3dagster:
Oh right... because both onSelect and onSideScroll would get called for the same item on initial render... Back to the drawing board

web/src/features/menu/list/index.tsx
```diff
@@ -50,6 +50,7 @@ const ListMenu: React.FC = () => {
   };
 
   const moveMenu = (e: React.KeyboardEvent<HTMLDivElement>) => {
+    if (firstRenderRef.current) firstRenderRef.current = false
```
LukeWasTakenn, line 53: Issue with this is that if the first button is a normal button the `onSelect` event won't fire.

LukeWasTakenn (changes requested):

LukeWasTakenn:
@jag3dagster I've pushed a possible fix so it *should* hopefully work as intended now, would appreciate it if you could test it as well.

jag3dagster:
So if I'm seeing that change correctly, you only want onSelect to fire on initial load/render?

LukeWasTakenn:
> So if I'm seeing that change correctly, you only want onSelect to fire on initial load/render?

Yeah that's the idea, it's also how it worked so far 🤷

jag3dagster:
Gotcha. I misunderstood them when I made that addition to onSelect. This should work as described now, I'll double-check when I get to my desk.

jag3dagster:
Yeup, just verified, this is working as described @LukeWasTakenn . I'm only seeing onSelected called on initial load / render and see everything else working on menu interactions.

LukeWasTakenn:
nice, thanks for the PR 👍

PR #174, opened 2022-12-14
feat(interface): Markdown wrapper, notify/context

davidxd33:
Added markdown support for context menu titles, item titles/descriptions, and notification wrapper description.


benjamin-ortizc:
good implementation dude

PR #180, opened 2022-12-17
scaleformMovie import

TheFarFantomas:


TheFarFantomas:
If you merge. I´ll write add this to docs

Vumono:
Why not like this in resource/streaming/client.lua
```lua
---Load a scaleform movie. When called from a thread, it will yield until it has loaded.
---@param scaleformName string
---@param timeout number? Number of ticks to wait for the scaleform movie to load. Default is 500.
---@return string? scaleformName

function lib.requestScaleformMovie(scaleformName, timeout)
	if HasScaleformMovieLoaded(scaleformName) then return scaleformName end
	return request(RequestScaleformMovie, HasScaleformMovieLoaded, 'scaleformName', scaleformName, timeout)
end```

TheFarFantomas:
I don´t know what you mean :D

PR #184, opened 2022-12-21
(feature) + Added Sound to Context Menu Events

itsHyper:


LukeWasTakenn:
Not everyone wants sounds so perhaps this should be either a convar or a per client setting

PR #200, opened 2023-01-14
fix(getNearbyVehicles): return correct coords

Aaron-Downham:
Currently it is returning the coords passed to the function, instead of the coords for each vehicle, this resolves that issue.


TheFantomas:
Practically, a commit was created that fixes several things, including this one. So you can probably conclude this :D #201

TheFantomas:
image

Issue #202, opened 2023-01-15
[Suggestion] Optional context search feature


Demigod916:
In a old context menu i made i had a feature where i could optionally put a search bar at the top of the menu so i could search/filter the menu buttons based on text content they contained. 

For example if you wanted to use a context menu at a car dealership, you could type the name of the car into the search bar and it would filter down the list as you type to be able to find it more easily in the massive list of cars.

LukeWasTakenn:
None of the menu's were designed with the idea to have a enormous amount of buttons in them in the first place, I don't really see this as something that makes a whole lot sense as for vehicle dealer resources you should really make a custom UI rather than use the context menu.

Moozdzn:
I added a search feature to nh-context and just finished adding a search feature to ox_lib context menu. Its pretty  simple but i can make a pull request for you to take a look at.

Demigod916:
Thats fair i guess. I can add it myself to my server, but just figured maybe others would find it useful.

LukeWasTakenn:
If you wish to implement it yourself you can look at Moozdzn's PR (#203) but it's not something we'll be adding first party support for.

PR #203, opened 2023-01-16
feat(interface/context): Search feature

Moozdzn:
This adds the feature requested at (https://github.com/overextended/ox_lib/issues/202). Like I mentioned in the conversation I'm creating this pull request from my changes. My experience with React is limited so there could a better way to achieve this.  
image

My use case is the context menu is used in the garage resource to spawn vehicles. As time went by the number of vehicles a player owns grew to the point where they requested a search bar to find a specific vehicle.

### DIFF
--- a/web/src/features/dev/debug/context.ts
+++ b/web/src/features/dev/debug/context.ts
@@ -8,6 +8,11 @@ export const debugContext = () => {
       data: {
         title: 'Vehicle garage',
         options: [
+          { 
+            type: "search", 
+            placeholder: "Search...",
+            icon: 'magnifying-glass',
+          },
           { title: 'Empty button' },
           {
             title: 'Example button',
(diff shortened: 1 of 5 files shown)


thelindat:
It’s a context menu; it’s not supposed to have infinite entries and be feature-rich. Pretty sure this will just cause issues with v3 also.

Moozdzn:
I understand that, but it is a feature I needed and I created a pull request since it was requested by another user. I leave to you to decide if its worth merging.

Moozdzn:
I don't know if context menu is finished on v3 but I can also take a look and add the feature there.

davidxd33:
> It’s a context menu; it’s not supposed to have infinite entries and be feature-rich. Pretty sure this will just cause issues with v3 also.

I can see this being useful for context menus with a lot of entries. Personally I have a few menus that has about 2 pages worth of entries for selecting food at a restaurant. Sometimes it's hard finding a particular item.

thelindat:
> I can see this being useful for context menus with a lot of entries.

> It’s a context menu; it’s not supposed to have infinite entries and be feature-rich

Other examples of context menus

image
image

LukeWasTakenn (commented):
It's just not something that we feel is to be used with a context menu.

PR #204, opened 2023-01-18
fix/feat(setClipboard): Allow use of setClipboard while menu is active

Moozdzn:
This could be a bug fix or a new feature depending on initial intentions.  FocusTrap prevents using lib.setClipboard while the menu is active. This commit fixes it by toggling the FocusTrap `paused` prop (don't know exactly on what it differs from the `active` prop) when lib.setClipboard is called while the menu is visible.


LukeWasTakenn (changes requested):
Doesn't seem to be working?

I have a command that opens a menu from the docs and another that runs `lib.setClipboard` with some value and the value isn't being copied when the menu is visible but is when the menu is closed.

PR #205, opened 2023-01-20
Adding support for vehicle deformations

Kiminaze:
... when using https://github.com/Kiminaze/VehicleDeformation

Just proposing this as an addition to this project.

A user requested me adding this into his garage script (which already uses ox_lib), so I thought, why not add it into it directly for everyone.


TheFantomas:
I doubt ox will want to add anything that requires third party resources...

PR #206, opened 2023-01-22
New Field types for V3 (Time and Radio Buttons)

masonschafercodes:
Don't know if this is necessarily wanted in v3, but I added it just to have more options on the date / time input

Edit:

Adding radio buttons here as well


masonschafercodes:
Decided that work is a fantasy... changes have been made and should be all good to go

LukeWasTakenn (commented):
LGTM overall, but since this returns a unix timestamp I'm assuming there's a date attached to it as well, is the date always the same or?

LukeWasTakenn:
> LGTM overall, but since this returns a unix timestamp I'm assuming there's a date attached to it as well, is the date always the same or?

Oh it seems to return the current date with the input time

LukeWasTakenn (approved):

PR #208, opened 2023-01-23
fix/feat(progressbar): Better animation cancelling

JnKTechstuff:
Changes:
- Stops setting the data.anim to a boolean value because we aren't using the same if statement for cancelling the tasks
- If animation then it will cancel only the animation played by the progress bar
- If scenario then it will clear ped tasks like before

This should make it so people can do things like eat sitting down and after a progress bar animation it won't make you stand up.


resource/interface/client/progress.lua
```diff
@@ -143,6 +135,13 @@ local function startProgress(data)
             end
         end
     end
+    
+    if data.anim then
```
mahanmoulaei, line 139: The ```if``` should check for ```data.anim?.dict``` instead
The ```elseif``` also should check for ```data.anim?.scenario```

Issue #178, opened 2022-12-16
[Suggestion] Additional modal options


PoorLatency:
**Is your feature request related to a problem? Please describe.**
A newer version of `modal` or extended version of `alertDialog` via https://mantine.dev/core/modal/ with more customization.

**Describe the solution you'd like**
```lua
-- data: table
-- header: string
-- content: string
-- size: 'xs' or 'sm' or 'md' or 'lg' or 'xl' or `auto` (optional - default is md)
-- centered: boolean (optional)
-- overflow: boolean (optional)
-- fullscreen: boolean (optional)


local modal = lib.displayModal({
	header = "text",
	content = 'General Kenobi  \n Markdown support!',
	size = "md",
	centered = false,
	overflow = false,
})
```

**Additional context**
Let me know if my request is closed due to it being v2. I'd like to see if this is possible.

LukeWasTakenn:
Can you clarify whether you want to see these options on just the alert dialog modal or the input dialog modal as well?

PoorLatency:
> Can you clarify whether you want to see these options on just the alert dialog modal or the input dialog modal as well?

Your choice is fine.

PR #213, opened 2023-01-31
feat(web/alert): new options for alert dialog

masonschafercodes:
In regards to this suggestion: #178

Added:
- size prop
- overflow prop

Updated:
- npm package types to reflect the newly added ones
- lua types to also reflect the newly added ones 

seem that the centered prop is already added as an option to the alert dialog.


LukeWasTakenn (approved):
lgtm

PR #217, opened 2023-02-04
feature: progress bar in context menu's metadata

szymczakovv:
On each metadata item you can set progressbar also you can still set value (not required) like this;
```
{
    ["label"]: "Body",
    ["value"]: "55%",
    ["progress"]: 55
}
```
or just like this;
```
{
    ["label"]: "Oil",
    ["progress"]: 11
}
```

new feat ox_lib

### DIFF
--- a/web/src/features/menu/context/components/ContextButton.tsx
+++ b/web/src/features/menu/context/components/ContextButton.tsx
@@ -90,10 +90,16 @@ const ContextButton: React.FC<{
         <HoverCard.Dropdown className={classes.dropdown}>
           {button.image && <Image src={button.image} />}
           {Array.isArray(button.metadata) ? (
-            button.metadata.map((metadata: string | { label: string; value: any }, index: number) => (
-              <Text key={`context-metadata-${index}`}>
-                {typeof metadata === 'string' ? `${metadata}` : `${metadata.label}: ${metadata.value}`}
-              </Text>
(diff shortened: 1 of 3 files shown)


web/src/features/menu/context/components/ContextButton.tsx
```diff
@@ -90,10 +90,16 @@ const ContextButton: React.FC<{
...
-                {typeof metadata === 'string' ? `${metadata}` : `${metadata.label}: ${metadata.value}`}
-              </Text>
+            button.metadata.map((metadata: string | { label: string; value?: any; progress?: number }, index: number) => (
+              <>
+                <Text key={`context-metadata-${index}`}>
+                  {typeof metadata === 'string' ? `${metadata}` : `${metadata.label}: ${metadata?.value ?? ""}`}
+                </Text>
+
+                {typeof metadata === 'object' && metadata?.progress && (
+                  <Progress value={metadata.progress || 0} size="sm" color={button.colorScheme || 'dark.3'} />
```
LukeWasTakenn, line 100: This needs to be a strict undefined check as 0 is false in JS so if `metadata.progress` was 0 this would display `0` instead of an empty progress bar.

LukeWasTakenn (changes requested):

LukeWasTakenn (approved):

PR #218, opened 2023-02-06
feat(web/context): mouse enter and mouse leave events

masonschafercodes:
does what it says. Adds two new things to context options for capturing mouse enter and mouse leave events on the button.

### DIFF
--- a/package/client/resource/interface/context.ts
+++ b/package/client/resource/interface/context.ts
@@ -6,6 +6,8 @@ interface ContextMenuItem {
   icon?: IconName | [IconPrefix, IconName];
   iconColor?: string;
   onSelect?: (args: any) => void;
+  onEnter?: (args: any) => void;
+  onLeave?: (args: any) => void;
   arrow?: boolean;
   description?: string;
   metadata?: string | { [key: string]: any } | string[];
--- a/package/client/resource/interface/context.ts
+++ b/package/client/resource/interface/context.ts
@@ -6,8 +6,7 @@ interface ContextMenuItem {
   icon?: IconName | [IconPrefix, IconName];
   iconColor?: string;
   onSelect?: (args: any) => void;
-  onEnter?: (args: any) => void;
-  onLeave?: (args: any) => void;
+  onHover?: (hoverState: boolean, args: any) => void;
   arrow?: boolean;
   description?: string;
   metadata?: string | { [key: string]: any } | string[];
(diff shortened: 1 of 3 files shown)


package/client/resource/interface/context.ts
```diff
@@ -6,6 +6,8 @@ interface ContextMenuItem {
   icon?: IconName | [IconPrefix, IconName];
   iconColor?: string;
   onSelect?: (args: any) => void;
+  onEnter?: (args: any) => void;
+  onLeave?: (args: any) => void;
```
LukeWasTakenn, line 10: These should probably be one function, something like:

```lua
onHover = function(entered)
    if entered then
        -- Do stuff when mouse enters
    else
        -- Do stuff when mouse leaves
    end
end

File: resource/interface/client/context.lua [hunk 2]
```diff
@@ -115,7 +117,36 @@ RegisterNUICallback('clickContext', function(id, cb)
...
 
-RegisterNUICallback('closeContext', closeContext)
+RegisterNUICallback('onEnter', function(id, cb)
+    cb(1)
 
+    if math.type(tonumber(id)) == 'float' then
+        id = math.tointeger(id)
+    elseif tonumber(id) then
+        id += 1
+    end
```
LukeWasTakenn, line 127: This conversion can go into it's own function I think, to avoid repeating it 3 times.

web/src/features/menu/context/components/ContextButton.tsx
```diff
@@ -12,6 +12,14 @@ const clickContext = (id: string) => {
...
 };
 
+const onEnterContext = (id: string) => {
+  fetchNui('onEnter', id);
+};
+
+const onLeaveContext = (id: string) => {
+  fetchNui('onLeave', id);
+};
+
```
LukeWasTakenn, line 22: These can go inline, same with `clickContext`.

LukeWasTakenn (changes requested):

masonschafercodes:
Let me change these things and test out to see if the leave event is triggered onClose.

masonschafercodes:
Seems that I get the same result when testing the it for this as well. If `ESC` is pressed there is no callback made for the `onMouseLeave`. Is this something we should simulate by triggering the callback when the Modal closes (only if the `onHover` options is even present)?

masonschafercodes:
could be good to keep the currently hovered button and then `onClose` we check to see if there is an active hovered button and if so then we trigger the callback for it, then clear the currently hovered state.

masonschafercodes:
That commit should cover the following cases:
- run the `onHover` function when the user clicks on an option that has an `onHover` attached to it
- run the `onHover` function when the user was hovering over a option with a `onHover` and closes the context menu

LukeWasTakenn (approved):

PR #224, opened 2023-02-17
Add help text support for command arguments

Mkeefeus:
Allows for use of an additional `:` after the arg type where you can specify help text for individual arguments

### DIFF
--- a/imports/addCommand/server.lua
+++ b/imports/addCommand/server.lua
@@ -1,106 +1,32 @@
-local commands = {}
-
-SetTimeout(1000, function()
-	TriggerClientEvent('chat:addSuggestions', -1, commands)
-end)
-
-AddEventHandler('playerJoining', function(source)
-	TriggerClientEvent('chat:addSuggestions', source, commands)
-end)
-
-local function chatSuggestion(name, parameters, help)
-	local params = {}
-
-	if parameters then
-		for i = 1, #parameters do
-			local arg, argType, helpText = string.strsplit(':', parameters[i])
-			argType = argType ~= '' and argType or nil
-
-			if argType and argType:sub(0, 1) == '?' then
-				argType = argType:sub(2, #argType)
-			end
-
-			params[i] = {
-				name = arg,
-				help = (helpText and argType and ("%s (%s)"):format(helpText, argType)) or argType or helpText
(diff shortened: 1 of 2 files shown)


Mkeefeus:
My initial thought was to use a table but I didn't want to break backwards compatibility. Something like this maybe? https://gist.github.com/Mkeefeus/4e71e157ba9f36c6380155e267adcb03

thelindat:
I already wrote up a replacement last night but need to test the backwards compatibility (would like to keep the addCommand name).. just waiting for FiveM to bother connecting.

thelindat:
```
BREAKING CHANGE: First parameter now holds command name(s).
Second parameter is properties, containing (optional) values for help,
restricted, and params.

Restricted is an ace permission (or array) i.e. 'group.admin', otherwise
'true' locks it behind standard command restrictions (command.commandName).
Setting it as 'false' or leaving it nil permits all users to trigger it.

Params is a table with the name, type (optional) and help text (optional).
The same types work as before (number/string) with the addition of playerId,
and are now more strictly checked against input.

The old syntax is available with lib.__addCommand, and partial compatibility
is implemented - but commands should be updated for v3.0.
```

## Old syntax (deprecated but handled for compatibility)
```lua
lib.addCommand('group.admin', 'testa', function(source, args)
	print(json.encode(args, {indent=true}))
end)

lib.addCommand('group.admin', {'testb'}, function(source, args)
	print(json.encode(args, {indent=true}))
end, {'target:number', 'item:string', 'count:number', 'metatype'})
```

## New syntax
```lua
lib.addCommand('testc', {
    help = 'Gives an item to a player',
    params = {
        {
            name = 'target',
            type = 'playerId',
            help = 'Target player\'s server id', },
        {
            name = 'item',
            type = 'string',
            help = 'Name of the item to give',
        },
        {
            name = 'count',
            type = 'number',
            help = 'Amount of the item to give, or blank to give 1',
            optional = true,
        },
        {
            name = 'metatype',
            help = 'Sets the item\'s "metadata.type"',
            optional = true,
        },
    },
    restricted = 'group.admin'
}, function(source, args, raw)
	print(json.encode(args, {indent=true}))
end)
```

PR #228, opened 2023-02-18
feat(web/radial): Add 'more' button

Moozdzn:
Prevents menu overflowing when there are too many items registered to a menu. Allows a maximum of 6 items (currently hardcoded) before showing a 'more' button to access the other items

Before (extreme example):
image

After:
image


LukeWasTakenn (changes requested):
I like the idea but the implementation for this seems quite weird,

I don't see why there needs to be a new state for the menu index.

Making opening a sub menu through the more button would probably be a better solution.

Moozdzn:
I tried a pure lua implementation but I felt it was more work. Since it should be all one menu for the user, there was a lot to keep track under hood regarding add/removeRadialItem and re-registering the same menu. I can look at my implementation and see if there is something I can refactor

LukeWasTakenn:
Implemented as pagination rather than separate menus with commit https://github.com/overextended/ox_lib/commit/7ffa12af4f2fec5e92a09269fed481907f3db5fc

PR #229, opened 2023-02-18
feat(web/radial): Add event and serverEvent option to item click

Moozdzn:
Allows the same flexibility of choice that already exists in context menu

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -8,7 +8,10 @@ local currentRadial = nil
 ---@field icon string
 ---@field label string
 ---@field menu? string
----@field onSelect? function
+---@field onSelect? fun(args: any)
+---@field event? string
+---@field serverEvent? string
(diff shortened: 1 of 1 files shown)


LukeWasTakenn:
Just use `onSelect` to run the events.

Moozdzn:
Any reason for this feature to be present in context menu but not the radial?

thelindat:
Stuff like this should trigger a specific action - if an event is wanted it can be fired after that. People tend to have weird design choices when given more flexibility, such as needing an "args" table on _everything_ even though it's a bunch of pre-defined _static_ data that could be defined locally in your resource, wherever it gets used.

If we could go back and remove support for these things we would, but we've tried that before and the amount of complaints is astounding.

Moozdzn:
I wondered if it was something like that but decided to make a PR just in case. Fair enough, thank you for the explanation

PR #230, opened 2023-02-18
feat: Support image Tag in buttons

renzuzu:
in my server. ox_lib is heavily used.
most of sample use cases for this is.
Food image,
Ingredients,
Vehicle Parts,
etc..

i believe font awesome is not enough for some use cases.

i did suggest this on v3 pending PR.
but it seems gets ignored.

here i am gonna try to push this PR.

to use this.
icon must be not existing in the options.

if both icon and image is in options, nothing will appear.

image in metadata will still appear normaly.

Sample:

image
image

sample usage:
```
				image = 'nui://ox_inventory/web/images/engine.png',
```

### DIFF
--- a/package/client/resource/interface/context.ts
+++ b/package/client/resource/interface/context.ts
@@ -6,7 +6,7 @@ interface ContextMenuItem {
   description?: string;
   arrow?: boolean;
   image?: string;
-  icon?: IconName | [IconPrefix, IconName];
+  icon?: IconName | [IconPrefix, IconName] | string;
   iconColor?: string;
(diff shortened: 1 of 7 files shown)


LukeWasTakenn (changes requested):
Images should be the same size as the icons, currently they are way too big.

renzuzu:
image
image

its now comparable to icons

LukeWasTakenn (approved):
Made the images the same size as icons to avoid creating awkward spacing.

Also moved the image to the `icon` prop instead.

Issue #234, opened 2023-02-20
[Bug] Radialmenu: leftover goback button when calling refreshItems in a nested submenu


Moozdzn:
**Describe the bug**
When insided a nested submenu, updating global radialmenu will cause the radial to open the global radialmenu but show a leftover go back icon in the center that does nothing instead of the `X` icon to close the radial.

**To Reproduce**
1. Register the following Menus/Items
```lua
lib.registerRadial({
  id = 'police_frisk',
  items = {
    {
      label = 'Frisk',
      icon = 'hand'
    },
    {
      label = 'Frisk',
      icon = 'hand'
    },
    {
      label = 'Frisk',
      icon = 'hand'
    },
  }
})

lib.registerRadial({
  id = 'police_handcuff',
  items = {
    {
      label = 'Handcuff',
      icon = 'handcuffs',
    },
    {
      label = 'Handcuff',
      icon = 'handcuffs',
    },
    {
      label = 'Handcuff',
      icon = 'handcuffs',
    },
  }
})

lib.registerRadial({
  id = 'police_menu',
  items = {
    {
      id = 'dandcuff',
      label = 'Handcuff',
      icon = 'handcuffs',
      menu = 'police_handcuff'
    },
    {
      id = 'frisk',
      label = 'Frisk',
      icon = 'hand',
      menu = 'police_frisk'
    },
    {
      label = 'Fingerprint',
      icon = 'fingerprint'
    },
    {
      label = 'Jail',
      icon = 'bus'
    },
    {
      label = 'Search',
      icon = 'magnifying-glass',
      onSelect = function()
        print('Search')
      end
    }
  }
})

lib.addRadialItem({
  {
    id = 'police',
    label = 'Police',
    icon = 'shield-halved',
    menu = 'police_menu'
  },
  {
    id = 'police2',
    label = 'Police 2',
    icon = 'shield-halved',
    menu = 'police_menu'
  },
  {
    id = 'police3',
    label = 'Police 3',
    icon = 'shield-halved',
    menu = 'police_menu'
  },
  
  {
    id = 'police4',
    label = 'Police 4',
    icon = 'shield-halved',
    menu = 'police_menu'
  },
  
  {
    id = 'police5',
    label = 'Police5',
    icon = 'shield-halved',
    menu = 'police_menu'
  },
  {
    id = 'garage_access',
    icon = 'warehouse',
    label = 'Garage',
    onSelect = function()
      print('Garage')
    end
  },
})

[...]

Issue #233, opened 2023-02-20
[Bug] Refresh global menu when in submenu


Moozdzn:
**Describe the bug**
If the player is in a submenu and the global radialmenu gets refreshed the UI will also update, removing the player from the current menu and the options will no longer match.

**To Reproduce**
1. copy and run the `Usage Example` from the docs
2. open radial menu and select the police menu
3. leave the mouse hovered over the handcuffs option
4. leave the neaby area to remove the garage option
5. the radial will go back to the global radial and clicking police should print `handcuffs`

**Expected behavior**
Open the police menu

**Screenshots**
[Video](https://streamable.com/y9gbn3)

PR #235, opened 2023-02-20
feat(radial): More button implemented in lua

Moozdzn:
Same feature as https://github.com/overextended/ox_lib/pull/228 but implemented in lua. Leaves the Ui only to display data and not setup extra menus.

Fixes https://github.com/overextended/ox_lib/issues/233
Fixes https://github.com/overextended/ox_lib/issues/234

Draft pull request because it may need some refactoring and just wanted to show to get some opinions

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -188,8 +188,10 @@ local function removeRadialItem(id)
             if item.id == id then
                 local lastGlobalMenuIndex = #globalMenus
                 local lastGlobalMenu = globalMenus[lastGlobalMenuIndex]
+
                 local lastIndex = #lastGlobalMenu
                 local lastItem = lastGlobalMenu[lastIndex]
+
                 item.id = lastItem.id
                 item.icon = lastItem.icon
                 item.label = lastItem.label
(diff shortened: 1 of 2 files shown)


Moozdzn:
As far as I know it doesn't have any bugs. It also wasn't supposed to be such a major rework, but I think it was necessary to implement this feature properly.

thelindat:
Rather than just spending time on posting two issues https://github.com/overextended/ox_lib/pull/228 https://github.com/overextended/ox_lib/pull/235 to do the same thing in different ways, wait for a response. We've already discussed the "more button" stuff internally but Luke has been busy.

Moozdzn:
Luke on the other pr responded with

> Opening a sub menu through the more button would probably be a better solution.

This requires a lua implementation and the reason for this pr , I just left the other one opened as a second alternative

LukeWasTakenn:
I'm not looking for a whole rewrite of the radial menu's Lua.

But yes it seems handling the menu in JS would be *the* better approach..?

I still need to find some to dedicate to it and play around with it.

Moozdzn:
It wasn't suppose to be one it sort of happened once I started making changes. I'll leave the other PR open for the time being.

Issue #92, opened 2022-09-11
[Suggestion] Some functionalities for input dialogs.


teidova21:
# Some functionalities for input dialogs.
### Live input custom-check with a boolean function
Adding a 'check' property to entries, a function taking in parameter the value and returning boolean of value validity.
```lua
function lockerNumberCheck(value)
    return string.len(value) == 4 and tonumber([value](s:sub(1, 1))) == nil
end
local input = lib.inputDialog('Identité', {
   { type = "input", label = "Locker ID", placeholder = "A000", check = lockerNumberCheck},
})
```
On field value change, call the function and, if invalid, change the background color of the field to a red bg color, and block the 'Confirm' button.

### min/max for number inputs too
Adding a 'min' and 'max' property, like on sliders, to number type entries, upon leaving focus on field or after using increment / decrement arrow, if it's < min, set value to min, if it's >max set value to max.
```lua
local input = lib.inputDialog('Gentleness test', {
    { type = "number", label = "Number counter", default = 12 }
})
```

### Pre-filled values
Adding a 'value' property to entries to define a pre-filled value. Useful when editing old values for example.
```lua
local input = lib.inputDialog('Gentleness test', {
  { type = "input", label = "Locker ID", placeholder = "A000", value = "B158" },
  { type = 'select', label = 'Cutest animal', value = 'meow_hater', options = {
  { value = 'kittens', label = 'Kittens' },
  { value = 'cats', label = 'Cats' },
  { value = 'meow_hater', label = 'I\'m idiot i don't like cats'},
  }}
})
```

### Date input type
Adding a 'date' input type when you need date input (i.e. character creation) :D
**Sub functionalities :**
Max / min date like on sliders too
```lua
local input = lib.inputDialog('Gentleness test', {
  { type = "date", label = "Date of birth", value = "B158" },
})
```

### Required fields
Adding a 'required' optional boolean attribute to all inputs, blocking the 'Confirm' button if it's empty (and add a * need the field name ?)
```lua
local input = lib.inputDialog('Gentleness test', {
  { type = "date", label = "Date of birth", required = true },
})
```

### Cancel button toggling
Adding an option for disabling option to leave menu without pressing 'Confirm', useful on menu that shouldn't be left unanswered like character creation eventually or stuff like this
I dont really know how to implement it tho
maybe a third optional argument as an object called 'options' for non field-related options
```lua
lib.inputDialog(heading, rows, options)
local input = lib.inputDialog('Gentleness test', {
[...]

PR #245, opened 2023-02-24
feat(input) : Support onChange callbacks

renzuzu:
not sure if this is the proper way to implement:

4th parameter in Input Dialog supports cb that returns any changed value.

Return Data is:
image

Supports only: Input, Slider, Number.
Usage:

```
        local input = lib.inputDialog('Police locker', {
		{ type = "slider", label = "Ignition Timing", min = -0.5, max = 1.5 , step = 0.001, default = 1.0},
		{ type = "number", label = "Camber Alignment", default = 1 },
		{ type = "input", label = "Gear Ratio", default = '3.251' },
		}, {}, function(data)
		print(json.encode(data, {indent = true}),'data')
	end)
```

This will be usefull on some cases that needs to see the real time change before applying.

### DIFF
--- a/resource/interface/client/input.lua
+++ b/resource/interface/client/input.lua
@@ -1,5 +1,4 @@
 local input
-
 ---@class InputDialogRowProps
 ---@field type 'input' | 'number' | 'checkbox' | 'select' | 'slider' | 'multi-select' | 'date' | 'date-range' | 'time' | 'textarea'
 ---@field label string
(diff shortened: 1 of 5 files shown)


LukeWasTakenn:
I don't really see the need for this?

Input dialog is mean to be a semi-simplistic submit form, if you need to run validation on the data you can do it when it's submitted.

Also sending a callback *every* time a key is pressed is really bad.

renzuzu:
> I don't really see the need for this?

i was trying to convert my other resource to ox_lib but the form needs to passed data everytime its change with its visual. ex. stancer wheel

> Also sending a callback _every_ time a key is pressed is really bad.

i was actually thinking the same too, to fix this theres a workaround like what you did in other fetchUI with catch and settimeout function. or use onchangeEnd.

> Input dialog is mean to be a semi-simplistic submit form, if you need to run validation on the data you can do it when it's submitted.

right, its just i think this feature is missing on the forms and it could be useful for other use cases too.

Moozdzn:
> Live input custom-check with a boolean function https://github.com/overextended/ox_lib/issues/92

Something like this is a valid use case.
To prevent sending a callback every time, one can trigger it only after the input loses focus

Issue #242, opened 2023-02-23
[Suggestion] Remember previous radial menu page when going back


Moozdzn:
**Is your feature request related to a problem? Please describe.**
Currently when going back to the previous menu it opens of the first page, even if the option was in one of the 'more' menu

**Describe the solution you'd like**
[What happens now](https://streamable.com/vz402n)
I think it would make sense when leaving the menu with the 'Frisk' options, the prev menu should be the one with 'Police 8' instead of having to click the more button again

LukeWasTakenn:
Should be resolved in 709d512...?

Could you double check it works as intended?

Moozdzn:
Don't think so, give me a minute to give you some steps so you can test it yourself

Moozdzn:
```lua
lib.registerRadial({
    id = "pet_interactions",
    items = {
        {
            icon = "dog",
            label = "call_pet",
            onSelect = function()
                TriggerEvent("pets:client:callPet")
            end
        },
        {
            icon = "hand",
            label = "dismiss_pet",
            onSelect = function()
                TriggerEvent("pets:client:dismissPet")
            end
        },
        {
            icon = "bone",
            label = "feed_pet",
            onSelect = function()
                TriggerEvent("pets:client:feedPet")
            end
        },
        {
            icon = "thumbs-up",
            label = "come",
            onSelect = function()
                TriggerEvent("pets:client:petFollow")
            end
        },
        {
            icon = "thumbs-down",
            label = "stay",
            onSelect = function()
                TriggerEvent("pets:client:petUnfollow")
            end
        },
        {
            icon = "car",
            label = "command_enter_car",
            onSelect = function()
                TriggerEvent("pets:client:enterCarPet")
            end
        },
        {
            icon = "user-shield",
            label = "toggle_defense_mode",
            onSelect = function()
                TriggerEvent("pets:client:toggleDefendMode")
            end
        },
        {
            icon = "ban",
            label = "cancel_command",
            onSelect = function()
                TriggerEvent("pets:client:resetPet")
            end
        },
        {
            icon = "hand-point-down",
            label = "order_sit",
            onSelect = function()
                TriggerEvent("pets:client:sitPet")
            end
        },
        {
            icon = "hand-holding",
            label = "order_lay_down",
            onSelect = function()
                TriggerEvent("pets:client:layPet")
            end
        },
        {
            icon = "exclamation",
            label = "order_bark",
            onSelect = function()
                TriggerEvent("pets:client:barkPet")
[...]

PR #247, opened 2023-02-26
fix(radial): go back to correct more page when going back from sub

Moozdzn:
Fixes https://github.com/overextended/ox_lib/issues/242

Since items lack an id, I'm storing the item label on `menuHistory` alongside the id. Also added the global radial menu to the `menuHistory` with id = nil.

added an option label parameter to `showRadial`  that gets passed to the NUI and used to find the page where that item will be


Moozdzn:
On second thought it may make more sense to store the `menu` field instead of the label

PR #249, opened 2023-02-26
Add a shouldClose option to radial item.

azawat7:
If shouldClose is set to false, the radial menu won't close when clicking an item. By default it is set to true.


eprimex:
Personally i would prefer it being by default false instead of true

PR #250, opened 2023-02-27
Add custom colors

Scullyy:
Also submitted a PR on the docs for this also


LukeWasTakenn:
We're not going to be adding custom colors named after other organizations, and especially not named or based on servers.

PR #257, opened 2023-02-27
feat(radial): Move default key to a convar

Moozdzn:
Saw someone on discord mentioning 'Z' key was conflicting with other resources that use it. Kept the same name convention as ox_target

Can also make a pr to the docs if this gets accepted

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -256,7 +256,7 @@ end)
 lib.addKeybind({
     name = 'ox_lib-radial',
     description = 'Open radial menu',
-    defaultKey = 'z',
+    defaultKey = GetConvar('ox_lib:defaultRadialkey', 'z'),
     onPressed = function()
         if isOpen then
             return lib.hideRadial()


thelindat:
The problem with adding a convar for default key is people assume it will magically and retroactively apply to everyone on the server.

If you’ve been on another server or the server you’re on used Z, joining a server where it isn’t Z doesn’t apply. If your first server uses P and you go to another server where it’s different, you’ll still use P.


It’s a constant headache and something I’ve been dealing with “issues” for ox inventory with for two years. Every key is capable of conflicting somehow somewhere (F2 is vMenu noclip for example), so changing the default achieves nothing when people can remap it themselves.

Moozdzn:
Yeah I completely understand.
 This was more to keep resource configuration in line with other ox resources where you don’t need to modify it in any way.

thelindat:
> keep resource configuration in line with other ox resources where you don’t need to modify it in any way

I'm pretty much at a point where I don't _want_ to support people making modifications and having certain settings hard-set rather than configurable, as most don't know what they're doing and make far too many weird assumptions and decisions as a result.

PR #261, opened 2023-03-01
fix(interface): Add table<string, string> to icon typings

Moozdzn:
`['far', 'heart']` is also a valid icon prop

I swear I read this on the docs, maybe on v2?

I can also update the docs if this gets merged

### DIFF
--- a/resource/interface/client/main.lua
+++ b/resource/interface/client/main.lua
@@ -1,3 +1,5 @@
+---@alias IconProp 'fas' | 'far' | 'fal' | 'fat' | 'fad' | 'fab' | 'fak' | 'fass'
+
 local keepInput = IsNuiFocusKeepingInput()
 
 function lib.setNuiFocus(allowInput, disableCursor)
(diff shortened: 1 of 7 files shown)


resource/interface/client/main.lua
```diff
@@ -1,3 +1,5 @@
+---@alias IconProp 'fas' | 'far' | 'fal' | 'fat' | 'fad' | 'fab' | 'fak' | 'fass'
```
LukeWasTakenn, line 1: Only `fas`, `far`, and `fab` are loaded into UI.

LukeWasTakenn (changes requested):

resource/interface/client/main.lua
Moozdzn: Without the others it will throw a warning to those who use fontawesome pro (like me).

Moozdzn: They are also all the IconPrefixes defined in https://github.com/FortAwesome/Font-Awesome/blob/6.x/js-packages/%40fortawesome/fontawesome-common-types/index.d.ts

LukeWasTakenn: Yeah that's true, you're right.

LukeWasTakenn (approved):

PR #264, opened 2023-03-05
feat(streaming): RequestWeaponAsset

Moozdzn:
Adds [RequestWeaponAsset](https://docs.fivem.net/natives/?_0x5443438F033E29C3) to the lib.

Added p1 as 31 and p2 as 0 for the lua version.

Regarding the npm package I used the generic function so p1 and p2 weren't included but was able to use [CreateWeaponObject](https://docs.fivem.net/natives/?_0x9541D3CF0D398F36) and [ShootSingleBulletBetweenCoords](https://docs.fivem.net/natives/?_0x867654CBC7606F2C) without issues when testing.

### DIFF
--- a/imports/requestWeaponAsset/client.lua
+++ b/imports/requestWeaponAsset/client.lua
@@ -1,16 +1,29 @@
 ---Load a weapon asset. When called from a thread, it will yield until it has loaded.
 ---@param weaponHash string | number
 ---@param timeout number? Number of ticks to wait for the asset to load. Default is 500.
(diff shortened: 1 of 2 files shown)


thelindat:
I’ll need to check but off the top of my head the parameters are related to animations and components; should probably be optional.

Moozdzn:
I used the ones the docs said were most common, but it can be changed to allow those parameters to be passed and if not default to 31 and 0

thelindat:
If you want this merged, needs the parameters added.

PR #268, opened 2023-03-11
Hidden Context Button

Axel2222:
Adding option to hide the button from the context menu using the param "hidden"

### DIFF
--- a/resource/interface/client/context.lua
+++ b/resource/interface/client/context.lua
@@ -11,6 +11,7 @@ local openContextMenu = nil
 ---@field description? string
 ---@field metadata? string | { [string]: any } | string[]
 ---@field disabled? boolean
+---@field hidden? boolean
 ---@field event? string
 ---@field serverEvent? string
 ---@field args? any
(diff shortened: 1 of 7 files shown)


Moozdzn:
What is the use case for this? Can you simply not add that option to the menu?
To toggle this state you have to re-register the whole menu again, no?

Axel2222:
For example, if you have one context menu with 10 options and you need to show only 3 of them, you can use hidden instead of using 'if' for every option. And the code will look much prettier.

thelindat:
Prettier is subjective, and you still have to register the menu again while including the options you want hidden; seems a bit pointless.

Moozdzn:
How do you determine if the option is supposed to be hidden or not without some type of condition

LukeWasTakenn:
I myself really don't need the point of this, definitely don't see it justifying the code complexity that comes with it, either don't add the button in at all or use the `disabled` prop.

Axel2222:
> Prettier is subjective, and you still have to register the menu again while including the options you want hidden; seems a bit pointless.

Don't you have to register the menu again if you want to disable the option?

thelindat:
> Don't you have to register the menu again if you want to disable the option?

Yes. And to make options "hidden" you can just not have those options in the first place 🤯

PR #272, opened 2023-03-13
Add clearAllRadialItems

Mkeefeus:
I used this for our conversion from qb-radialmenu to clear the menu then only add a "call for help" option or similar. Figured it could be useful in other niche situations

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -182,6 +182,15 @@ function lib.removeRadialItem(id)
     refreshRadial(id)
 end
 
+---Removes all items from the global radial menu.
+function lib.clearAllRadialItems()
+    if isOpen then
+        lib.hideRadial()
+    end
+
+    table.wipe(menuItems)
+end
+
 RegisterNUICallback('radialClick', function(index, cb)
     cb(1)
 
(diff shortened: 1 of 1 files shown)


resource/interface/client/radial.lua
```diff
@@ -182,6 +182,15 @@ function lib.removeRadialItem(id)
     refreshRadial(id)
 end
 
+---Removes all items from the global radial menu.
+function lib.clearAllRadialItems()
```
LukeWasTakenn, line 186: I think `clearRadialItems` would be a bit better than `clearAllRadialItems`

resource/interface/client/radial.lua
```diff
@@ -182,6 +182,15 @@ function lib.removeRadialItem(id)
...
+    if isOpen then
+        lib.hideRadial()
+    end
```
LukeWasTakenn, line 189: Why not do a check for `not currentRadial` as well and update the menu instead of closing it?

LukeWasTakenn (changes requested):
Could you also make a PR to the docs for this function?

Mkeefeus:
Requested updates made and the PR for docs is open

LukeWasTakenn (approved):

PR #284, opened 2023-03-25
feat(interface/radial): don't show more button in pages with 8 slices

D4isDAVID:
Previously, menu pages with 8 buttons/slices would move the 8th button to the next page and replace it with a `More...` button.

This PR changes the functionality so that a `More...` button is added only if there are *more* than 8 buttons in the current page.

### DIFF
--- a/web/src/features/menu/radial/index.tsx
+++ b/web/src/features/menu/radial/index.tsx
@@ -52,8 +52,8 @@ const useStyles = createStyles((theme) => ({
   },
 }));
 
-// PAGE_ITEMS + 1 = More... button
-const PAGE_ITEMS = 7;
+// includes More... button
+const PAGE_ITEMS = 8;
 
 const degToRad = (deg: number) => deg * (Math.PI / 180);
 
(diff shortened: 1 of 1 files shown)


LukeWasTakenn:
Any specific reasoning for this PR? what's wrong with 7 + 1 slices with the way it is currently?

D4isDAVID:
When you have just 8 buttons they can fit in a single page, but currently they don't and instead a More button appears.
I saw someone complain say something about this in the Qbox Discord and I thought too that a new page for a single button seems kinda redundant.

jag3dagster:
I agree. Setting up a whole new page for 1 extra button when you could just have all 8 on a single page feels weird UX-wise

LukeWasTakenn:
Would this not be the same issue if you have 9 items in total as well...?

D4isDAVID:
That's why I changed more than just the button count

D4isDAVID:
Perhaps I misunderstood, when you have 9 items then a More button is appropriate since 9 items can't fit in a menu which should contain 8 items.

LukeWasTakenn:
Right I see. I'll review next time I'm working on lib.

LukeWasTakenn (approved):
Seems to be working well

PR #286, opened 2023-04-02
feat(interface/input): minLength & maxLength option

FilipeCuco:


LukeWasTakenn (approved):
Can you make a PR to the docs that reflect these changes? It should also be noted that `minLength` requires the `required` prop to be set to true.

FilipeCuco:
yes, I will do that now

PR #294, opened 2023-04-11
feat(logger/loki): Allow users to specify protocol for requests

TheiLLeniumStudios:
This is helpful for simple environments which don't have SSL enabled
- Defaults to `https`

### DIFF
--- a/imports/logger/server.lua
+++ b/imports/logger/server.lua
@@ -3,6 +3,32 @@ local hostname = GetConvar('sv_projectName', 'fxserver')
 local buffer
 local bufferSize = 0
 
+local b = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/"
+
+local function base64encode(data)
+    return ((data:gsub(".", function(x)
+        local r, b = "", x:byte()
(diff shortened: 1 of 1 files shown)


ahcenezdh:
You should allow user to not use a password too, since loki can be used without a password

thelindat:
endpoint should rather support the full address (prepending https if no protocol is included).

TheiLLeniumStudios:
@thelindat should I then remove the user and password vars as well? Since those need to be part of the URL? Users will then need to provide it as part of the URL as well. Because if I don't, then I'll have to split the URL up just to extract the protocol from it and recreate the full URL

thelindat:
Nothing wrong with splitting up the string but right now the convars are a bit of a mess and certain things are hardcoded which shouldn't be. Well, whatever works for you - I don't and haven't used Loki; logger itself needs a rewrite so right now easiest solutions are best.

TheiLLeniumStudios:
Well, another way would be to use a Basic Auth authorization header instead of appending the credentials in the URL. I will play around with it and update the P

TheiLLeniumStudios:
@thelindat can you take a look again please?

PR #302, opened 2023-04-20
fix: clipboard copy multi line now

Andyyy7666:


Andyyy7666:
Basically input is only one line so it won’t copy multiple lines it will put everything in one line so this will make everything like it should and put it in multiple lines and then copy it because textarea is an area for text and multiple lines that way you can now copy multiple lines and not only one line which is really useful. Basically explain this massive pull request.

PR #303, opened 2023-04-20
Fix for table.concat nil

userMacieG:
While using creator after creating poly zone, I found out that the last point is always missing `zCoord` if not confirmed.


userMacieG:
Why?

PR #309, opened 2023-04-25
[Bug Fix] Fix issue with menu item focus not updating in ListItem component

ATG-Github:
This pull request addresses an issue where the menu item selector/focus was not updating correctly in the ListItem component. The changes ensure that the focus updates as expected when navigating through the menu items.

The changes made are as follows:
1. Update the `ref` prop type in the `Props` interface to use `React.Ref<HTMLDivElement>`.
```
interface Props {
  item: MenuItem;
  index: number;
  scrollIndex: number;
  checked: boolean;
  ref: React.Ref<HTMLDivElement>;
}
```
2. Update the `ref` assignment inside the `Box` component in the `ListItem` component.
```
<Box
  tabIndex={index}
  className={classes.buttonContainer}
  key={`item-${index}`}
  ref={ref}
>
```

These changes fix the issue where the selector/focus was not updating as expected when navigating through the menu items, leading to a better user experience and improved accessibility.


ATG-Github:
This issue occurred with a larger menu. I would use the arrow keys to change selection, and the only thing that would change was the description at the bottom of the menu. This solves said bug. Interestingly, there were zero errors to, so this was a doozy to figure out.

jag3dagster:
How big of a menu is a 'larger' menu? Give actual repro steps or example code where this occurs because I have never experienced it before and I've had menus of 10-15 items without issue

userMacieG:
I used menu in esx_lscustom and sometimes this actually happen. I tried to trigger that issue to record that, but it's really random.
No error, nothing.

PR #311, opened 2023-04-27
radial opened / closed event

JonasDev17:
An event that gets triggered when the menu is opened or closed.
Could also be named onOpen / onClose or onShow / onHide, opened / hidden....

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -126,6 +126,7 @@ function lib.hideRadial()
 
     isOpen = false
     currentRadial = nil
+    TriggerEvent('ox_lib:radial:closed')
 end
 
 ---Registers an item or array of items in the global radial menu.
@@ -307,6 +308,8 @@ lib.addKeybind({
         if #menuItems == 0 or IsNuiFocused() or IsPauseMenuActive() then return end
 
         isOpen = true
+            
+        TriggerEvent('ox_lib:radial:opened')
 
         SendNUIMessage({
             action = 'openRadialMenu',


TonybynMp4:
OMG that is revolutionnary thank you so much for this upgrade that will make the society a better place!

ARSSANTO:
it will help us to add radial menu items from other script with condition

thelindat:
Because brigading is what gets features accepted. 😒
Explain the actual use-case (without saying "well qb does it!"). Like hell I'm using that event name too.

> it will help us to add radial menu items from other script with condition

??????

TonybynMp4:
> Because brigading is what gets features accepted. 😒 Explain the actual use-case (without saying "well qb does it!"). Like hell I'm using that event name too.

Basically without an event you have to loop while you're inside the zone to chekc if you got in or out of a car and stuff, that allows to not loop and just check when the menu is actually being opened

ARSSANTO:
> Because brigading is what gets features accepted. 😒
> Explain the actual use-case (without saying "well qb does it!"). Like hell I'm using that event name too.
> 
> > it will help us to add radial menu items from other script with condition
> 
> ??????

you are smort already get why this pr opend 

if not ok so in qb there is a trigger event like onradialmenuopen  which are being registered so many other creator script  + jdevs qb garages 
therefor we can register the same event on client to manipulate sets of menu items through function #sorry for bad English its not my 1st language

thelindat:
> Basically in our case without an event you have to loop while you're inside the zone to check if you got in or out of a car and stuff, that allows to not loop and just check when the menu is actually being opened, wich can be useful for other cases too

Triggering an event with 100 listeners for "radialopened" is worse than a slow loop anyway. Can also accomplish a lot using zones onEnter/onExit and cache.vehicle/lib.onCache.

> if not ok so in qb there is a trigger event like onradialmenuopen which are being registered so many other creator script + jdevs qb garages

> Explain the actual use-case (without saying "well qb does it!").

thelindat:
Closed since you're all just being weirdos spamming comments "wow so good add this" and reacting with 👀 rather than giving me any reasonable use for this event.

TonybynMp4:
> Closed since you're all just being weirdos spamming comments "wow so good add this" and reacting with 👀 rather than giving me any reasonable use for this event.

I mean after your last comment i had nothing more to say you provided a good solution lol

PR #314, opened 2023-04-30
tweak(server/logger): backwards compatibility for loki convar

daZepelin:


imports/logger/server.lua
```diff
@@ -115,7 +115,8 @@ end
 
 if service == 'loki' then
     local lokiUser = GetConvar('loki:user', '')
-    local lokiPassword = GetConvar('loki:password', '')
+    local lokiKey = GetConvar('loki:key', '')
+    local lokiPassword = GetConvar('loki:password', lokiKey ~= '' and lokiKey or '')
```
thelindat, line 119: uhh
```lua
local lokiPassword = GetConvar('loki:password', GetConvar('loki:key', ''))
```
?

daZepelin: Yea, fair enough, overthough it somehow myself.

PR #316, opened 2023-04-30
feat(radial): Support newlines

jag3dagster:


jag3dagster:
Can make the newline token `  \n` instead of just `\n` to match the markdown from elsewhere in lib.

jag3dagster:
Now using `  \n` to match markdown from elsewhere in lib.

I also emulated some different resolutions via browser testing and those looked fine to me as well. Since 'em' is just the size of the font it shouldn't be a problem.

LukeWasTakenn (approved):
Seems to be working well, thanks!

PR #317, opened 2023-05-01
fix(progressbar): spamming cancel issue

mikigoalie:
Issue:
Spamming starting and cancelling progress bar would sometimes result in prog bar returning true eventhough all are cancelled. I believe it is due to wrong condition race.

Repro: Bind prog bar to a key and spam cancel, if youre quick enough, you can cancel prog bar and start another one before the first one finished.

It's minor bug but I'd like to see this patched. I tested it on my seatbelt script and worked fine.

Linden mentioned it doesnt make sense or whatever, I didn't mess too much with the code, I only removed local variable and moved progress = nil below cancel check.

### DIFF
--- a/resource/interface/client/progress.lua
+++ b/resource/interface/client/progress.lua
@@ -147,13 +147,13 @@ local function startProgress(data)
     end
 
     playerState.invBusy = false
-    local cancel = progress == false
-    progress = nil
 
-    if cancel then
+    if progress == false then
         SendNUIMessage({ action = 'progressCancel' })
         return false
     end
+
+    progress = nil
 
     return true
 end
--- a/resource/interface/client/progress.lua
+++ b/resource/interface/client/progress.lua
@@ -150,12 +150,9 @@ local function startProgress(data)
 
     if progress == false then
         SendNUIMessage({ action = 'progressCancel' })
-        return false
     end
 
-    progress = nil
-
-    return true
+    return progress == true
 end
 
 ---@param data ProgressProps


mikigoalie:
PTYM
By default, I can start two prog bars and cancel previous one, resulting in no prog bar being shown in UI and also unable to be cancelled


I wanted to keep the original code without messing too much into it not to break anything else, however you can probably also remove progress = nil

thelindat:
Your repro still makes no sense, and no matter how fast I spam cancel (or even automate canceling it) they will not start until the previous bar has ended.

Are you even running the latest version / built the UI after this commit? https://github.com/overextended/ox_lib/commit/672b638ba747f1bd95cfe44910f040cf1b1dbc00

mikigoalie:
You fixed it in this commit ^^ https://github.com/overextended/ox_lib/commit/878763867a758b4fdc717b1aaee28a11e65824ce 

After latest changes I believe it's patched. Thanks

thelindat:
> You fixed it in this commit ^^ [8787638](https://github.com/overextended/ox_lib/commit/878763867a758b4fdc717b1aaee28a11e65824ce)

Like I said

> The does add a transition / increase the delay between progress bars which is probably better UX, but it isn't fixing anything

Nothing you said happens happened.

mikigoalie:
Yeah, what can I say. Unfortunate you couldn't get it to repro, but I've had issues with it if progress bar was running through a keybind.

I believe the issue was here https://github.com/overextended/ox_lib/commit/878763867a758b4fdc717b1aaee28a11e65824ce#diff-8046dfd65c488341a4bd5cbb122992b2d1766e977a5ce5830a8350e3e4a03b16L151

Next time I will try to provide better repro and more detailed info related to the bug, thanks though.

PR #319, opened 2023-05-02
tweak(server/cron): getNextTime param to return time for any day

daZepelin:
This allows `getNextTime` to return an absolute next time (ignoring if it's today or not) to be returned. Can be useful in some scenarios.

### DIFF
--- a/imports/cron/server.lua
+++ b/imports/cron/server.lua
@@ -12,10 +12,11 @@ currentDate.sec = 0
 ---@field isActive boolean
 ---@field id number
 ---@field getNextTime function
+---@field getAbsoluteNextTime function
 ---@field debug? boolean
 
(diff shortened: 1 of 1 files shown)


thelindat:
You'd want to handle for year as well

daZepelin:
> You'd want to handle for year as well

Thanks for pointing that out, added that

thelindat:
When I tested this a ~week ago it broke existing behaviour. iirc, it just assumes it needs to get the next year if you've gone past a set day of the month (and week even?). You need to test a bunch of different tasks and confirm the behaviour matches before and after if you want this sort of change in.

I get there are some use cases where checking the next time is useful but it was only intended for internal use and done in the easiest/lazy manner.

PR #331, opened 2023-05-17
Budget way to disable radial menu when esc

Sh0rtyyy:


LukeWasTakenn:
There's already plenty of ways to hide the radial menu, you can toggle it with the same key, you can right click out of it and you can press the X button in the middle.

I don't see why we would need a 4th way of hiding the radial menu?

Sh0rtyyy:
Until know I know only about the "X" way. 
And why ? 
Because ESC is most common way to close something, and I think a lot of people are pressing esc for close. And opening the map and have radial open is a bot very good looking.

PR #336, opened 2023-05-18
feat(client/cache): adds cached 'dead' status of a player

0xDEMXN:


jag3dagster:
Would this even work? Don't most cores revive the player and have them play a 'downed' animation instead of leaving them actually dead?

0xDEMXN:
> Would this even work? Don't most cores revive the player and have them play a 'downed' animation instead of leaving them actually dead?

I mean, do we care?
Not everyone use that death system and imo this could be just useful for many people.

jag3dagster:
> I mean, do we care?

Yes because we'll have to deal with all of the support questions about "Why isn't this working?"

I don't think it's a bad feature, I just think the implementation doesn't cover the majority of cases, or at least handles them in a confusing way.

0xDEMXN:
> > I mean, do we care?
> 
> Yes because we'll have to deal with all of the support questions about "Why isn't this working?"
> 
> I don't think it's a bad feature, I just think the implementation doesn't cover the majority of cases, or at least handles them in a confusing way.

Well if the player is revived and put down he's not dead so the "dead" value will be correct anyway.

0xDEMXN:
Also I don't see a lot of servers using that system (at least in my country).

thelindat:
No. Death systems are more complicated than setting a dead value, so this is useless with every known framework anyway.

PR #338, opened 2023-05-23
update(math): group digits

mikigoalie:
thought it'd be a good idea to include this in math lib

### DIFF
--- a/imports/math/shared.lua
+++ b/imports/math/shared.lua
@@ -90,9 +90,6 @@ function math.tohex(n, upper)
     return upper and n:upper() or n
 end
 
-return lib.math
-
-
 ---Converts input number into grouped digits with optional divider and currency char
 ---@param number number
 ---@param currencyChar? string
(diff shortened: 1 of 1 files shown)


mikigoalie:
Credits for the formatting goes https://stackoverflow.com/questions/10989788/format-integer-in-lua

thelindat:
The number type-check is san issue, if say you want to group `"$10000"`.
Then there's the issue of math.floor, because `1000.347` becomes `"1,000"`. Of course, removing it results in bad formatting as `"1,000.,347"`.

This one from lua-users is more robust https://lua-users.org/wiki/FormattingNumbers.
```lua
function comma_value(n) -- credit http://richard.warburton.it
	local left,num,right = string.match(n,'^([^%d]*%d)(%d*)(.-)$')
	return left..(num:reverse():gsub('(%d%d%d)','%1,'):reverse())..right
end
```

mikigoalie:
hawaii weather pr

thelindat:
> hawaii weather pr

????

imports/math/shared.lua
```diff
@@ -90,4 +90,14 @@ function math.tohex(n, upper)
...
 
-return lib.math
+---Converts input number into grouped digits
+---@param number number
+---@param useDot? boolean
+---@return string
+function math.groupdigits(number, useDot) -- credit http://richard.warburton.it
+    local divider = useDot and '.' or ','
+    local left,num,right = string.match(number,'^([^%d]*%d)(%d*)(.-)$')
+    return left..(num:reverse():gsub('(%d%d%d)','%1' .. divider):reverse())..right
```
thelindat, line 100: might as well be `divider` or `seperator` as the arg (not useDot), and then just do `divider or ','`.

PR #341, opened 2023-05-26
feat: added splunk hec logging

IdrisDose:
Back at it again with a logging integration. This time it's for those who have access to Splunk. I know it's incredibly small. In fact, one person is using it, but I'd rather commit this now than ignore the logger file manually when merging updates.

Feature: Add Splunk as an option for receiving logs.
Reasoning: Why not? Adds an extra option for logging output.

Will there be documentation? If you are using this level of logging software, you should be able to read and follow previous examples.

### DIFF
--- a/imports/logger/server.lua
+++ b/imports/logger/server.lua
@@ -232,7 +232,7 @@ if service == 'splunk' then
         ['Authorization'] = 'Splunk ' .. splunkToken
     }
 
-    if not endpoint:find(startingPattern) then
+    if not splunkEndpoint:find(startingPattern) then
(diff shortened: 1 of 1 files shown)


thelindat:
Not going to merge.

Aside from further bulk in this already messy logger, Splunk looks even less desirable for FiveM than the other options and definitely not suited as someone's introduction to logging services.

Await the modular change so you can easily make and share your own, however I'd like to keep officially supported options cheap and relatively simple to start with. The more services we support, the more questions we get about using them - and I barely uses Datadog and Grafana as-is.

PR #351, opened 2023-06-11
feat(resource/vehicleProperties): add support for custom xenon colors

ahcenezdh:
Add support for custom xenon colors (based on ESX.GetVehicleProperties) and use RemoveVehicleWindow instead of SmashVehicleWindow


ahcenezdh:
Oh ok i didn't know !

PR #360, opened 2023-06-25
fix(web/input): Blocks form submission with unclamped numbers

RIOT1K:
Now if someone enters a number that is lower or higher than allowed and presses enter instead of the send button, it will send unclamped number.


LukeWasTakenn:
Seems I've only committed the change to the date input field type and not the number one ☹️

PR #361, opened 2023-06-26
tweak(server/logger) remove text formatting from hostname

jellyton69:
Remove the text formatting codes commonly used in server names when setting the hostname to be used in the logger.

### DIFF
--- a/imports/logger/server.lua
+++ b/imports/logger/server.lua
@@ -15,7 +15,7 @@ local function removeColorCodes(str)
     return str
 end
 
-local hostname = removeColorCodes(GetConvar('sv_projectName', 'fxserver'))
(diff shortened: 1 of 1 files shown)


thelindat:
This should probably be using `sv_hostname` to be honest?

jellyton69:
For datadog specifically, the hostname is supposed to be set so it matches with the datadog agent's hostname so the logs link with the infrastructure metrics, so it would be better if we could specify the hostname manually with a convar for that purpose. But sv_hostname is normally the long form of the server name, at least in my experience, so the project name is fine in this case.

thelindat:
New convar to specify a logger hostname (not specific to datadog) and defaulting to the projectName is fine.

PR #364, opened 2023-07-04
- (FEAT) Added feature to cancel skillcheck

MatinHQ:
- (FEAT) Added feature to check is any active skillcheck

### DIFF
--- a/resource/interface/client/skillcheck.lua
+++ b/resource/interface/client/skillcheck.lua
@@ -34,7 +34,7 @@ end
 
 ---@return boolean
 function lib.skillCheckActive()
-    return (skillcheck and true) or false
+    return skillcheck ~= nil
 end
 
 RegisterNUICallback('skillCheckOver', function(success, cb)
--- a/resource/interface/client/skillcheck.lua
+++ b/resource/interface/client/skillcheck.lua
@@ -29,6 +29,7 @@ function lib.cancelSkillCheck()
 
     skillcheck:resolve(false)
     skillcheck = nil
+    SendNUIMessage({action = 'skillCheckCancel'})
 end
 
 ---@return boolean
--- a/resource/interface/client/skillcheck.lua
+++ b/resource/interface/client/skillcheck.lua
@@ -22,6 +22,20 @@ function lib.skillCheck(difficulty, inputs)
     return Citizen.Await(skillcheck)
 end
 
+function lib.cancelSkillCheck()
+    if not skillcheck then
+        error('No skillCheck is active')
+    end
+
+    skillcheck:resolve(false)
+    skillcheck = nil
+end
+
+---@return boolean
+function lib.skillCheckActive()
+    return (skillcheck and true) or false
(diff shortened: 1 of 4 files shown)


resource/interface/client/skillcheck.lua
```diff
@@ -22,6 +22,20 @@ function lib.skillCheck(difficulty, inputs)
...
+        error('No skillCheck is active')
+    end
+
+    skillcheck:resolve(false)
+    skillcheck = nil
+end
+
+---@return boolean
+function lib.skillCheckActive()
+    return (skillcheck and true) or false
```
BerkieBb, line 36: Return an actual boolean easier
```suggestion
    return skillcheck ~= nil
```

resource/interface/client/skillcheck.lua
```diff
@@ -22,6 +22,20 @@ function lib.skillCheck(difficulty, inputs)
     return Citizen.Await(skillcheck)
 end
 
+function lib.cancelSkillCheck()
+    if not skillcheck then
+        error('No skillCheck is active')
```
BerkieBb, line 27: Instead of making an error, I'd advise to just return false if it fails and true if it succeeded

BerkieBb:
Also, this doesn't actually cancel the UI part, only does the lua part

MatinHQ:
> Also, this doesn't actually cancel the UI part, only does the lua part

Fixed hiding UI part

resource/interface/client/skillcheck.lua
MatinHQ: > Instead of making an error, I'd advise to just return false if it fails and true if it succeeded

They did same for progress bar cancel I agree with what you saying but I am just following ox standards

resource/interface/client/skillcheck.lua
MatinHQ: fixed

LukeWasTakenn:
What's the actual useful use case for this?

MatinHQ:
> What's the actual useful use case for this?

I have a robbery system designed for my server. In this system, one person acts as the robber, brandishing a gun on a NPC. The robber needs to perform skill checks during the robbery process.

However, there's a crucial event implemented to enhance the realism and dynamics of the robbery. If the person holding the gun decides to remove it from the cashier . hands will go down and robbery will be automatically canceled.

This event has significant implications, allowing me to interrupt the checkpoint for the player who is actively robbing the cashier. Before implementing this feature, the robbing person could continue performing skill checks even after the robbery was canceled, leading to inconsistencies and immersion-breaking experiences."

LukeWasTakenn (changes requested):
Make the TS wrappers for these functions and PR the docs so that his can be merged.

MatinHQ:
Done ts wrapper is already there and pr has been made to docs (https://github.com/overextended/overextended.github.io/pull/142)

LukeWasTakenn (approved):

PR #369, opened 2023-07-11
Add debugColour and debug switcher to zones

BerkieBb:
Something small that I'd like to add for scripts to do stuff with, for example highlighting a zone.

Usage example:

```lua
local zone = lib.zones.box({...})

-- Maybe later in the code
zone:setDebug(true, vec4(255, 100, 100, 120))
```


imports/zones/client.lua
```diff
@@ -261,6 +258,21 @@ local function convertToVector(coords)
     return coords
 end
 
+local function setDebug(self, bool, colour)
+    if not bool and insideZones[self.id] then
+        insideZones[self.id] = nil
+    end
+
+    self.debugColour = bool and vec4(colour?.r or self.debugColour?.r or 255, colour?.g or self.debugColour?.g or 42, colour?.b or self.debugColour?.b or 24, colour?.a or self.debugColour?.a or 100) or nil
+    Zones[self.id].debugColour = self.debugColour
```
thelindat, line 267: As `Zones[self.id] == self` there shouldn't be a need to reassign this property.

imports/zones/client.lua
```diff
@@ -261,6 +258,21 @@ local function convertToVector(coords)
...
+
+    if bool and self.debug or not bool and not self.debug then return end
+
+    self.triangles = bool and (self.type == 'poly' and getTriangles(self.polygon) or self.type == 'box' and { mat(self.polygon[1], self.polygon[2], self.polygon[3]), mat(self.polygon[1], self.polygon[3], self.polygon[4]) }) or nil
+    self.debug = bool and (self.type == 'sphere' and debugSphere or debugPoly) or nil
+    Zones[self.id].debug = self.debug
```
thelindat, line 273: `Zones[self.id] == self`

imports/zones/client.lua
```diff
@@ -328,13 +340,16 @@ lib.zones = {
...
+        data.setDebug = setDebug
 
         if data.debug then
             data.debug = nil
 
             CreateThread(function()
+                data.debugColour = vec4(data.debugColour?.r or 255, data.debugColour?.b or 42, data.debugColour?.g or 24, data.debugColour?.a or 100)
                 data.triangles = getTriangles(data.polygon)
                 data.debug = debugPoly
             end)
```
thelindat, line 355: Should probably just call `self:setDebug(...)` to cut down on duplicate code (here and on the other zone types).

imports/zones/client.lua
```diff
@@ -4,6 +4,7 @@ local glm = require 'glm'
 ---@field id number
 ---@field coords vector3
 ---@field distance number
+---@field type string
```
thelindat, line 7: String literal type `---@field type 'poly' | 'sphere' | 'box'`.
Might need to mark debug as private and add the setDebug method too.

thelindat (changes requested):

imports/zones/client.lua
BerkieBb: Yeah I wasn't sure if it would update it

imports/zones/client.lua
BerkieBb: After marking debug as private
image

PR #372, opened 2023-07-17
Changes to input fields

daZepelin:
- fix: added missing disabled prop to checkbox
  - checkbox disabled prop used to not work

- feat: added precision prop for number input
  - without this, the number from input was always returned as integer, while in some cases float might be useful

### DIFF
--- a/web/src/features/dialog/components/fields/number.tsx
+++ b/web/src/features/dialog/components/fields/number.tsx
@@ -30,6 +30,7 @@ const NumberField: React.FC<Props> = (props) => {
       defaultValue={props.row.default}
       min={props.row.min}
       max={props.row.max}
+      precision={props.row.precision}
       disabled={props.row.disabled}
       icon={props.row.icon && <FontAwesomeIcon icon={props.row.icon} fixedWidth />}
       withAsterisk={props.row.required}
--- a/web/src/features/dev/debug/input.ts
+++ b/web/src/features/dev/debug/input.ts
@@ -41,6 +41,15 @@ export const debugInput = () => {
             icon: 'receipt',
           },
           {
+            type: 'number',
+            label: 'Price',
+            default: 6.5,
+            min: 0,
+            max: 10,
+            step: 0.5,
+            icon: 'price-tag',
+          },
+          {
             type: 'slider',
             label: 'Slide bar',
             min: 10,
--- a/web/src/features/dev/debug/input.ts
+++ b/web/src/features/dev/debug/input.ts
@@ -46,7 +46,6 @@ export const debugInput = () => {
             default: 6.5,
             min: 0,
             max: 10,
-            step: 0.5,
             icon: 'receipt',
           },
           {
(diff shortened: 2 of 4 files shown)


web/src/features/dialog/components/fields/number.tsx
```diff
@@ -30,6 +30,7 @@ const NumberField: React.FC<Props> = (props) => {
       defaultValue={props.row.default}
       min={props.row.min}
       max={props.row.max}
+      precision={props.row.precision}
       disabled={props.row.disabled}
```
LukeWasTakenn, line 34: The step prop is missing but is defined in the typings.

LukeWasTakenn (changes requested):

LukeWasTakenn (changes requested):
Why remove the step prop from the typings instead of passing it to the actual component?

daZepelin:
> Why remove the step prop from the typings instead of passing it to the actual component?

As I understand, the project is aiming to make it as simplistic as possible for usage and I am not entirely sure if this is a neccesary feature, as barely any people use the arrow buttons to increase or decrease the values. But I can surely add that if you think it will be needed, as myself I added it just by accident by testing

LukeWasTakenn:
> > Why remove the step prop from the typings instead of passing it to the actual component?
> 
> As I understand, the project is aiming to make it as simplistic as possible for usage and I am not entirely sure if this is a neccesary feature, as barely any people use the arrow buttons to increase or decrease the values. But I can surely add that if you think it will be needed, as myself I added it just by accident when testing

Sure I guess that makes sense, remove the step prop from the debug dialog and then this can be merged.

masonschafercodes:
> barely any people use the arrow buttons to increase or decrease the values

In my opinion this is the wrong way to approach simplicity, simplicity has never meant removing a feature rather making that feature approachable for the end user. Just cause **you** haven't seen people use the step feature, doesn't mean people don't use it.

jag3dagster:
> implicity has never meant removing a feature rather making that feature approachable for the end user

This is the core of an never ending debate. On one hand, you have the user's perspective of preference or accessibility, on the other hand you have the overhead of code maintenance and supporting various ways to accomplish the same task. Ultimately it is up to the maintainer of a code base to decide which side of 'simplicity' they want to lean into as it is on them in the long run to support things as a project continues to move forward.

daZepelin:
> > barely any people use the arrow buttons to increase or decrease the values
> 
> In my opinion this is the wrong way to approach simplicity, simplicity has never meant removing a feature rather making that feature approachable for the end user. Just cause **you** haven't seen people use the step feature, doesn't mean people don't use it.

This feature was never removed, it has never been added, I accidentally added the step prop when testing as I though it will make the input fields support the float numbers.
I do understand that some might use it, but don't see it as necessary as it supporting float numbers.
The mantine library (the one used for the inputs) has like 30 - 40 props on each input field type and I hornestly do not think that all of them should be implemented in ox_lib and as I saw lots of them not being implemented I'd choose not to add any unnecessary ones.

LukeWasTakenn (approved):

PR #377, opened 2023-07-25
feat(zones): new zone type: interior

Fancazista:
I was creating a resource using ox_lib zones and I was thinking, why do I need to create a poly zone or a box for an interior when I can use the native to verify that my player is inside of it?

Using the native [GetInteriorFromEntity](https://docs.fivem.net/natives/?_0x2107BA504071A6BB) I can get the interior id of the interior in which the player is located and thanks to it I can check if the player enters or leaves the the zone.

Using the native [GetInteriorAtCoords](https://docs.fivem.net/natives/?_0xB0F7F8663821D9C3) I can get the interior id at certain coordinates and using this thing I can check if a point is inside the zone by comparing the interior id obtained from the native with the interior id of the zone.

Unfortunately, on the **server side**, it is not possible to check if certain coordinates are part of a interior because there is no native to verify this / obtain the interior id at determinate coordinates.

### DIFF
--- a/imports/zones/client.lua
+++ b/imports/zones/client.lua
@@ -4,7 +4,7 @@ local glm = require 'glm'
 ---@field id number
 ---@field coords vector3
 ---@field distance number
----@field type 'poly' | 'sphere' | 'box'
+---@field type 'poly' | 'sphere' | 'box' | 'interior'
 ---@field debugColour vector4?
 ---@field setDebug fun(self: CZone)
 ---@field remove fun()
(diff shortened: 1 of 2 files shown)


thelindat:
Interiors don't make sense in the context of zones, and we ideally want to split code into a shared file to reduce a lot of duplicate code (making interiors again an issue without server support). Not that an interior check like this doesn't have merit.

We probably need a more generic module for handling onEnter/onExit/whileInside checks to use for points and zones, and then use that for interior-checks too. But not important nor a focus.

PR #383, opened 2023-07-31
add modColor and fix problem with primary color 1 set and secondary color not set

ItsEwen:


thelindat:
Has some issues and another PR looks to be solving it a bit better.

PR #391, opened 2023-08-06
fix(client/vehicleproperties): extras & fix(web): heading props

jellyton69:
Vehicle Properties: When using `SetVehicleAutoRepairDisabled`, you need to repair the vehicle for applied extras to be visible.

Interface: Fixed weird heading margins when using markdown heading props.
August06 - 03 24 - 2641

Sorry for 2 commits per pull, I couldn't figure out how to separate them.

### DIFF
--- a/resource/vehicleProperties/client.lua
+++ b/resource/vehicleProperties/client.lua
@@ -277,11 +277,13 @@ end
 ---@param props VehicleProperties
 ---@return boolean?
 function lib.setVehicleProperties(vehicle, props)
-    if not DoesEntityExist(vehicle) then error(("Unable to set vehicle properties for '%s' (entity does not exist)"):
-            format(vehicle))
+    if not DoesEntityExist(vehicle) then
+        error(("Unable to set vehicle properties for '%s' (entity does not exist)"):
+        format(vehicle))
     end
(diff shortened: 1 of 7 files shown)


thelindat:
Preferred unrelated commits go in separate PRs.

---

Users can just pass SetVehicleFixed themselves, or otherwise add an arg to fix the vehicle?

LukeWasTakenn (changes requested):
> Sorry for 2 commits per pull, I couldn't figure out how to separate them.

Cherry pick the commits onto separate branches then open a PR from each of the branches to master.

`markdowncomponents.tsx` file should follow the naming convention so far, `MarkdownComponents.tsx`

jellyton69:
> Cherry pick the commits onto separate branches then open a PR from each of the branches to master.

Alrighty I'll close this, fix the requested changes and reopen it with separate pulls.

jellyton69:
> Users can just use SetVehicleFixed themselves, or otherwise add an arg to fix the vehicle?

Would you rather it be a new value in the props table, or an extra param in the setVehicleProperties function?

thelindat:
Just a new function arg.

PR #392, opened 2023-08-06
fix(client/vehicleproperties): vehicle must be fixed for extras to apply visually

jellyton69:
When using SetVehicleAutoRepairDisabled, you need to repair the vehicle for the applied extras to be visible.

### DIFF
--- a/resource/vehicleProperties/client.lua
+++ b/resource/vehicleProperties/client.lua
@@ -277,11 +277,13 @@ end
 ---@param props VehicleProperties
 ---@return boolean?
 function lib.setVehicleProperties(vehicle, props)
-    if not DoesEntityExist(vehicle) then error(("Unable to set vehicle properties for '%s' (entity does not exist)"):
-            format(vehicle))
+    if not DoesEntityExist(vehicle) then
+        error(("Unable to set vehicle properties for '%s' (entity does not exist)"):
+        format(vehicle))
     end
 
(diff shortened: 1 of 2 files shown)


thelindat:
Not sure why extras are moved up? Need the TS types updated to reflect the new optional arg.

resource/vehicleProperties/client.lua
```diff
@@ -275,13 +275,16 @@ end
 
 ---@param vehicle number
 ---@param props VehicleProperties
+---@param fixVehicle? boolean
 ---@return boolean?
-function lib.setVehicleProperties(vehicle, props)
-    if not DoesEntityExist(vehicle) then error(("Unable to set vehicle properties for '%s' (entity does not exist)"):
-            format(vehicle))
+function lib.setVehicleProperties(vehicle, props, fixVehicle)
```
thelindat, line 280: Requires TS definition.

resource/vehicleProperties/client.lua
```diff
@@ -291,6 +294,12 @@ function lib.setVehicleProperties(vehicle, props)
     SetVehicleModKit(vehicle, 0)
     SetVehicleAutoRepairDisabled(vehicle, true)
 
+    if props.extras then
+        for id, disable in pairs(props.extras) do
+            SetVehicleExtra(vehicle, tonumber(id) --[[@as number]], disable == 1)
+        end
+    end
+
```
thelindat, line 302: Why is this moved up?

thelindat (changes requested):

PR #401, opened 2023-08-24
Add step property to NumberInput.

Lunar-Scripts:
Added a missing property for NumberInput.


LukeWasTakenn (commented):
Step really doesn't do much unless you're using the scroll wheel or increment/decrement controls to change the value so not really sure how useful this actually is?

Lunar-Scripts:
The default step value isn't suitable for all situations when using the increment/decrement controls.
I am sure some people will find this useful too.

LukeWasTakenn (approved):
Can you add the change to the docs as well?

Lunar-Scripts:
Sure, no problem: [https://github.com/overextended/overextended.github.io/pull/151](PR)

LukeWasTakenn (approved):

PR #411, opened 2023-09-08
Add returnString field to date and date-range type in input

BerkieBb:
returnString allows for a date to be returned as a string following the format if it is defined, otherwise returns to a default format.

### DIFF
--- a/package/client/resource/interface/input.ts
+++ b/package/client/resource/interface/input.ts
@@ -29,6 +29,7 @@ interface InputDialogRowProps {
   step?: number;
   required?: boolean;
   format?: string;
+  useFormat?: boolean;
   description?: string;
 }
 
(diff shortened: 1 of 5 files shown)


package/client/resource/interface/input.ts
```diff
@@ -29,6 +29,7 @@ interface InputDialogRowProps {
   step?: number;
   required?: boolean;
   format?: string;
+  useFormat?: boolean;
```
LukeWasTakenn, line 32: Perhaps rather than `useFormat` maybe something a bit more descriptive? i.e `returnString` or something alike?

LukeWasTakenn (approved):
I've removed the unnecessary state and refactored the code a bit, it should be ok now.

Can you make a PR to the docs?

BerkieBb:
Will do later

PR #421, opened 2023-09-18
feat(checkDependency): minimum version optional

Manason:
Made minimum version an optional parameter of checkVersion so that it can be used to check for the existence of a dependency, as an alternative to fxmanifest dependencies, which have possibly undesired side effects such as stopping a resource when one of its' dependencies is stopped.


thelindat:
This is weird; also - GetResourceState?

PR #425, opened 2023-09-21
feat: added lib.print to print different log levels to console

Manason:
- added convar 'ox:printlevel'
- can override the global setting on a per-resource basis with 'ox:printlevel:<resourceName>'
- supported levels are debug, info, warn, and error. Considered adding trace and severe, but think that might just be overkill
- Examples:

Client:
image

Server: Unfortunately duplicating the resource name seems necessary so that when chat forwards the print, the client can see the original resource.
image

Server forwarded print to client through Chat resource:
image

### DIFF
--- a/imports/print/shared.lua
+++ b/imports/print/shared.lua
@@ -16,7 +16,7 @@ local levelPrefixes = {
 }
 
 local resourcePrintLevel = printLevel[GetConvar('ox:printlevel:' .. cache.resource, GetConvar('ox:printlevel', 'info'))]
-local template = ('^5[%s] %s %s^7'):format(cache.resource)
+local template = ('^5[%s] %%s %%s^7'):format(cache.resource)
 local jsonOptions = { sort_keys = true, indent = true }
 
 ---Prints to console conditionally based on what ox:printlevel is.
@@ -25,7 +25,7 @@ local jsonOptions = { sort_keys = true, indent = true }
 ---@param pattern string
 ---@param ... any
 local function libPrint(level, pattern, ...)
-    if level < resourcePrintLevel then return end
+    if level > resourcePrintLevel then return end
 
     local formattedArgs = {}
     for i = 1,select('#', ...) do
--- a/imports/print/shared.lua
+++ b/imports/print/shared.lua
@@ -1,0 +1,37 @@
+local printLevelConvar = GetConvar('ox:printlevel', 'info')
+
+local PrintLevel = {
+    debug = {
+        level = 1,
+        prefix = '^6[DEBUG] ',
+    },
+    info = {
+        level = 2,
+        prefix = '^7[INFO] ',
+    },
+    warn = {
+        level = 3,
+        prefix = '^3[WARN] ',
+    },
+    error = {
+        level = 4,
+        prefix = '^1[ERROR] ',
+    },
+}
+
+---Prints to console conditionally based on what ox:printlevel is.
+---Any print with a level more severe will also print. If ox:printlevel is info, then warn and error prints will appear as well, but debug prints will not.
+---@param level 'debug' | 'info' | 'warn' | 'error'
+---@param message any
+function lib.print(level, message)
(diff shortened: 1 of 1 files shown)


imports/print/shared.lua
```diff
@@ -0,0 +1,44 @@
...
+---@param message any
+function Print(level, message)
+    message = type(message) == "string" and message or json.encode(message)
+    local printLevel = PrintLevel[level]
+    if printLevel.level < PrintLevel[printLevelConvar].level then return end
+
+    -- server prints are forwarded to the client from the chat resource, so this lets clients see the original resource.
+
+    local resourceName = IsDuplicityVersion() and '^5[' .. cache.resource .. '] ' or ''
+    print(string.format('^5%s%s%s^1', resourceName, printLevel.prefix, message))
```
thelindat, line 34:
```lua
-- top level
local template = ('^5[%s] %%s^1'):format(cache.resource)

print(template:format(message))
```

imports/print/shared.lua
```diff
@@ -0,0 +1,44 @@
...
+        prefix = '^1[ERROR] ',
+    },
+}
+
+---Prints to console conditionally based on what ox:printlevel is.
+---Any print with a level more severe will also print. If ox:printlevel is info, then warn and error prints will appear as well, but debug prints will not.
+---@param level 'debug' | 'info' | 'warn' | 'error'
+---@param message any
+function Print(level, message)
+    message = type(message) == "string" and message or json.encode(message)
```
thelindat, line 27: This should instead check for `'table'` and encode only then. Should pass options through as well (store the options as a local rather than making a new table each time).

```lua
local options = { sort_keys = true, indent = true }
```

thelindat (changes requested):

imports/print/shared.lua
```diff
@@ -0,0 +1,44 @@
+local printLevelConvar = GetConvar('ox:printlevel', 'info')
```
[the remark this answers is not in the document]
Manason: Are you saying you prefer the convar to be of type int? I figured the string format was easier for people to understand

imports/print/shared.lua
Manason: Doesn't cache.resource need to read on each invocation of lib.print?

imports/print/shared.lua
```diff
@@ -0,0 +1,44 @@
...
+    if printLevel.level < PrintLevel[printLevelConvar].level then return end
+
+    -- server prints are forwarded to the client from the chat resource, so this lets clients see the original resource.
+
+    local resourceName = IsDuplicityVersion() and '^5[' .. cache.resource .. '] ' or ''
+    print(string.format('^5%s%s%s^1', resourceName, printLevel.prefix, message))
+end
+
+lib.print = {
+    debug = function(message) Print('debug', message) end,
```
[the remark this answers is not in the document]
Manason: I prefer the former as it's faster to type and easier to search for. Happy to change it back to the latter though if you prefer.

imports/print/shared.lua
```diff
@@ -0,0 +1,44 @@
+local printLevelConvar = GetConvar('ox:printlevel', 'info')
+
+local PrintLevel = {
```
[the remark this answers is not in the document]
Manason: fixed. This was originally an enum.

imports/print/shared.lua
Manason: done.

imports/print/shared.lua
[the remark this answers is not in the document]
Manason: Oh, I guess you meant just dereference it here. Take another look after the refactor.

imports/print/shared.lua
Manason: added cache.resource again, forgot this was imported.

Manason:
> @Manason Would you mind adding this to the docs?

https://github.com/overextended/overextended.github.io/pull/154

PR #427, opened 2023-09-25
feat: added lib.registerNetEvent

Manason:
- surfacing this ox_core lib function for ox_lib to register net events that cannot be called from the same side of the network


Moozdzn:
As mentioned by linden in discord check source should suffice
`if source == '' then return end`
no need for a function call

Manason:
> As mentioned by linden in discord check source should suffice
> `if source == '' then return end`
> no need for a function call

done

PR #443, opened 2023-10-15
fix(require): plaster solution for escrowed scripts

FjamZoo:
Currently if any part of the script is escrowed (tested on a couple of scripts) the resourceSrc would return ?

This is really the definition of a plaster solution feel free to close if you have a better way of doing this thanks.


FjamZoo:
Best solution I had to fix this, it's become quite a issue lately idk if you guys have a better solution.

thelindat:
> idk if you guys have a better solution

don't use [ass]et protection

PR #445, opened 2023-10-22
Update menu.lua

Zaps6000:
Fixing a nil error for a native not on redm

### DIFF
--- a/resource/interface/client/menu.lua
+++ b/resource/interface/client/menu.lua
@@ -52,12 +52,15 @@ function lib.showMenu(id, startIndex)
     end
 
     if not openMenu then
+
         local control = cache.game == 'fivem' and 140 or 0xE30CD707
         CreateThread(function()
             while openMenu do
                 if openMenu.disableInput == nil or openMenu.disableInput then
                     DisablePlayerFiring(cache.playerId, true)
-                    HudWeaponWheelIgnoreSelection()
+                    if cache.game == 'fivem' then      
+                    HudWeaponWheelIgnoreSelection() -- Not a REDM native
+                        end
                     DisableControlAction(0, control, true)
                 end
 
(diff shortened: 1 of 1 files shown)


LukeWasTakenn (commented):
Is there not a similar or equivalent native for RedM?

Zaps6000:
> Is there not a similar or equivalent native for RedM?

If this is I couldnt find it I Just know that every time you open a menu it returns nil

Zaps6000:
I looked again and am pretty sure there isn't one unless it isn't on the docs

resource/interface/client/menu.lua
```diff
@@ -52,12 +52,15 @@ function lib.showMenu(id, startIndex)
     end
 
     if not openMenu then
+
```
LukeWasTakenn, line 55: Don't remove the space

resource/interface/client/menu.lua
```diff
@@ -52,12 +52,15 @@ function lib.showMenu(id, startIndex)
...
         local control = cache.game == 'fivem' and 140 or 0xE30CD707
         CreateThread(function()
             while openMenu do
                 if openMenu.disableInput == nil or openMenu.disableInput then
                     DisablePlayerFiring(cache.playerId, true)
-                    HudWeaponWheelIgnoreSelection()
+                    if cache.game == 'fivem' then      
+                    HudWeaponWheelIgnoreSelection() -- Not a REDM native
```
LukeWasTakenn, line 62: This should be indented

resource/interface/client/menu.lua
```diff
@@ -52,12 +52,15 @@ function lib.showMenu(id, startIndex)
...
+                        end
```
LukeWasTakenn, line 63: This should not be indented

LukeWasTakenn (changes requested):
Fix the weird formatting issues and then it should be good to to be merged.

Zaps6000:
> Fix the weird formatting issues and then it should be good to to be merged.

should be good

LukeWasTakenn (approved):

PR #448, opened 2023-10-29
feat(imports/addKeybind): Support for RedM

tomasalves8:
For some reason this was removed in [018112b](https://github.com/overextended/ox_lib/pull/162/commits/018112b345e31a29dbe58ff31c5223ad87a19d82). 
I believe this workaround should be kept until RegisterKeyMapping is implemented in RedM, so there is some compatibility with resources that use addKeybind.

[RegisterKeyMapping RedM](https://github.com/citizenfx/fivem/issues/1779)

### DIFF
--- a/imports/addKeybind/client.lua
+++ b/imports/addKeybind/client.lua
@@ -1,4 +1,46 @@
-if cache.game == 'redm' then return end
+if cache.game == 'redm' then
+    local keybinds = {}
+
+    local IsPauseMenuActive = IsPauseMenuActive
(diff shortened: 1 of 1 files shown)


thelindat:
> For some reason this was removed in [018112b](https://github.com/overextended/ox_lib/pull/162/commits/018112b345e31a29dbe58ff31c5223ad87a19d82).

Not removed, was never merged in the first place.

thelindat (changes requested):

PR #449, opened 2023-11-03
Proposal: Add source to loki logging for searchability in grafana dashboard

NietThijmen:


thelindat:
Would prefer playerId or sessionId or such, as well as adding the tag for dd too.

thelindat:
Note for why this is dead:
> adding the tag for dd too.

NietThijmen:
> adding the tag for dd too.
What do you mean by tagging DD?

PR #458, opened 2023-11-18
feat(imports): get entities on server

Andyyy7666:
This is needed on the server


Andyyy7666:
@LukeWasTakenn this is ready

thelindat:
I'd rather not just see the code duplicated everywhere, i.e. adding a function to replicate `GetGamePool` on the server and then making most of the modules shared.

Andyyy7666:
There would be a slight problem with a few functions such as the `getNearbyPlayers` one where you can include getting the player but on the server you'd want to get any player so that wouldn't make sense.

thelindat:
> making most of the modules shared

or otherwise ignoring client-specific args.

Andyyy7666:
Didn't merge getting player since that works differently on server

Moozdzn:
I think you can use lib.context instead of multiple IsDuplicityVersion

Andyyy7666:
> I think you can use lib.context instead of multiple IsDuplicityVersion

I don’t recall lib.context being a feature, can you give me an example of what you mean?

Moozdzn:
https://github.com/overextended/ox_lib/blob/6c40dd2889cbc7edb153fe4a4e98143bc44f4fbd/init.lua#L89

Andyyy7666:
Thank you I'll use that instead.

PR #463, opened 2023-11-29
feature (imports/locale): Read nested locales

choxen:
Allow's to nest, group locales
Usage
```locale('notifcations.something_went_wrong')```


thelindat:
Would be better to flatten the object when it's loaded instead.

PR #482, opened 2024-01-11
TS support for addCommand

ESK0:
I would like to mention in advance that I am completely new to TS and thanks to the transition of ox_core to TS I got the motivation to dive more into this and learn the language.

I looked at what is available in ox_lib for TS and addCommand came to me as one of the few must have functions that ox_lib currently lacks..

I am more than sure that this PR has flaws and I would appreciate any feedback and change requests

Tested with:
```
import { addCommand } from '@overextended/ox_lib/server'

addCommand('tscommand', (source: number, args: any, raw: string) => {
    console.log(`player ${source} executed comand (${raw})`)
}, {help: 'TypeScript addCommand', params: [{name: 'target', paramType: 'playerId', help: 'playerId or "me"', optional: true}], restricted: ['group.typescript', 'group.typescript2'] } )
```

### DIFF
--- a/package/server/resource/addCommand/index.ts
+++ b/package/server/resource/addCommand/index.ts
@@ -1,133 +1,140 @@
-import { addAce } from "../acl";
+import { addAce } from '../acl';
 
-type OxCommandArguments = Record<(string|number), string | number | boolean>;
+type OxCommandArguments = Record<string | number, string | number | boolean>;
 
 interface OxCommandParams {
-    name: string;
-    help?: string;
-    paramType?: 'number' | 'playerId' | 'string';
-    optional?: boolean;
+  name: string;
+  help?: string;
+  paramType?: 'number' | 'playerId' | 'string';
+  optional?: boolean;
 }
 
 interface OxCommandProperties {
-    name?: string;
-    help?: string;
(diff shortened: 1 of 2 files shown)


package/server/resource/addCommand/index.ts
```diff
@@ -0,0 +1,133 @@
...
+function parseArguments(source: number, args: OxCommandArguments, raw: string, params?: OxCommandParams[]): OxCommandArguments | undefined {
+    if (!params) {
+        return args;
+    }
+
+    let paramMissing = false
+
+    params.every((param, index) => {
+        let arg = args[index]
+        let value: any
```
LukeWasTakenn, line 40: there's no reason not to type this when we are assigning it later and expect it to be of certain type

LukeWasTakenn (changes requested):
A lot of uses of `let` when most of these variables can be `const`, should always use `const` unless you really *must* use `let`.

Also please use Prettier with the config that's in the root of the `package` dir and reformat your code with it.

ESK0:
Thanks for review! Ill try to solve all requests later today or tomorrow

ESK0:
Should be all good now... fingers crossed

PR #487, opened 2024-01-20
feat(package): support icon urls in radial menu

ktx-mega:
I like the radial menu.
I use it in one of my projects, and it lack the option to use non FA icons.

This PR is an example implementation to add support for this.
I added width and height parameters to adapt to the images. 

I also fixed the type in debug/radial.ts.

Documentation update can be found here : #https://github.com/overextended/overextended.github.io/pull/169

### DIFF
--- a/package/client/resource/interface/radial.ts
+++ b/package/client/resource/interface/radial.ts
@@ -3,9 +3,11 @@ import type { IconName, IconPrefix } from '@fortawesome/fontawesome-common-types
 type RadialItem = {
   id: string;
   label: string;
-  icon: IconName | [IconPrefix, IconName];
+  icon: IconName | [IconPrefix, IconName] | string;
   onSelect?: (currentMenu: string | null, itemIndex: number) => void | string;
   menu?: string;
+  iconWidth?: number;
+  iconHeight?: number;
 };
 
 export const addRadialItem = (items: RadialItem | RadialItem[]) => exports.ox_lib.addRadialItem(items);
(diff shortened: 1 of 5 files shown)


LukeWasTakenn:
What you are proposing are images though, they aren't icons, they don't scale like SVGs and they tend to be blurry.

Also the ability to change the height and width is just a disaster waiting to happen with someone releasing a resource using the radial menu.

ktx-mega:
In my use case, I generate the radial items dynamically based on the response of an http query. This query includes an image link that is used as an icon. It is not vectorial, but unless very much upscaled it is not blurry. From what I tested I didn't have any blurry results.

The ability to change the height and width is here because the image size and offset might differ in different cases and that allow to fix this. The test image I provided in the PR looks good with a width/height of 35, whereas the images I use in my project looks better with 75. 
I don't think this is a disaster waiting to happen since it is managed by the person writing the mod, so they should be responsible when using this feature. In the same way, setting 3 lines long label won't look good either, yet it is not a disaster.

ktx-mega:
Here's what it looks like for me, with a size of 75 : 
image

ktx-mega:
Also, this feature is implemented in the context menu, I reused the code to test the icon parameter from there. The only difference is that you can't change the size there.

thelindat (commented):
Should be fine? Might be good to limit size though.
Need type changes for ts package and luadocs.

ktx-mega:
Thanks for the feedback.

Sorry about the types, I don't use TS so I forgot the interface, and I'm not familiar with LUA, I didn't see the doc.
I clamped the size between 0 and 100. I'm not a fan of limiting upward because it is an arbitrary value but this should reduce the risks of disasters.

Did I miss anything else ?

thelindat (approved):

ktx-mega:
Up,
Luke, anything else I need to do ?

LukeWasTakenn (approved):

PR #491, opened 2024-02-01
feat(imports/waitFor): add function callback for timeouts

puttydotexe:
This feature change allows developers to execute a function if the waitFor times out. It refactors the variable `errMessage` to `onTimeout` to reflect the change and to prevent it from breaking old code. 

The default usage of waitFor that prints an error message if it times out.
```lua
  return lib.waitFor(function()
      -- ...
  end, "it timed out", 500)
```

The new alternative usage of waitFor that can either print an error message or call a function if it times out.
```lua
  return lib.waitFor(function()
      -- ...
  end, function()
      -- It timed out, execute the code here.
  end, 500)
```

These features have been tested and all appear to be working as intended with newly written code using this and old code using the `errMessage` variable.

### DIFF
--- a/imports/waitFor/shared.lua
+++ b/imports/waitFor/shared.lua
@@ -1,11 +1,11 @@
 ---Yields the current thread until a non-nil value is returned by the function.
 ---@generic T
 ---@param cb fun(): T?
----@param errMessage string?
+---@param onTimeout string | fun()?
 ---@param timeout number? Error out after `~x` ms. Defaults to 1000, unless set to `false`.
 ---@return T?
 ---@async
-function lib.waitFor(cb, errMessage, timeout)
+function lib.waitFor(cb, onTimeout, timeout)
     local value = cb()
 
     if value ~= nil then return value end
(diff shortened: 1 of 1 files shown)


puttydotexe:
I just wanted to prevent any misunderstanding, this doesn't remove/break the usage of the error message. It adds to it by checking if it was parsed a function or a string and then either excutes said function or parses the string onto the print error function it previously used.

thelindat:
fwiw you can just pcall it.

```lua
local success, resp = lib.waitFor(function()
      -- ...
end, "it timed out", 500)

if not success then
  -- ...
end
```

puttydotexe:
Yeah I've seen some people using that however I just thought this wraps it up nicely. I obviously have no issue if this is to be considered redundent and not worth including into ox_lib, however due to it being a library and libraries are normally used to make things just that little bit easier I thought it'd be a nice addition.

thelindat:
I just don't think there's much benefit and it doesn't really make things easier. `pcall` is basically equivalent to try/catch block, and requiring explicit error handling isn't out of the ordinary.

This still requires you to either explicitly call error inside your timeout function or otherwise check if the result of waitFor was `nil` and return so not much/any time is saved.

I also need to be a bit stricter on changes like these being implemented both Lua and TS when appropriate.

PR #499, opened 2024-02-11
feat(cache): add playername cache entry

nwvh:
This PR will probably be useless, as the player's name won't change until he rejoins with a different name.

Usage:

```lua
print("player name: ", cache.playername)
```

### DIFF
--- a/resource/cache/client.lua
+++ b/resource/cache/client.lua
@@ -17,6 +17,7 @@ local GetVehicleMaxNumberOfPassengers = GetVehicleMaxNumberOfPassengers
 local GetMount = GetMount
 local IsPedOnMount = IsPedOnMount
 local GetCurrentPedWeapon = GetCurrentPedWeapon
+local GetPlayerName = GetPlayerName
 
 CreateThread(function()
 	while true do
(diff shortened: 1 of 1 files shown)


resource/cache/client.lua
```diff
@@ -51,6 +52,10 @@ CreateThread(function()
 
 		cache:set('weapon', hasWeapon and currentWeapon or false)
 
+		local playername = GetPlayerName(PlayerId())
```
Mirrrrrow, line 55: why don't you directly use the cached playerId instead of PlayerId() and use camelCase

resource/cache/client.lua
```diff
@@ -51,6 +52,10 @@ CreateThread(function()
...
+
+		cache:set('playername', playername)
```
Mirrrrrow, line 57: here also use camelCase

Mirrrrrow (commented):
i would suggest this

resource/cache/client.lua
nwvh: you're right, also just realized I don't have to cache the player name in a loop as it doesn't change, I can just cache it once after the script has loaded

resource/cache/client.lua
nwvh: I pushed another commit. should be all good now, thanks for the heads up

nwvh:
Updated the commit, now using camelCase and `cache.playerId` instead of `PlayerId()` as @Mirrrrrow suggested. I also modified it so that the playername is not being cached in the loop, but only once, just like the cache.playerId.

[#499 ](https://github.com/overextended/ox_lib/commit/61a7a7e2ce7885bb927f900288cb4029d60df57a)

thelindat:
Unnecessary. I probably wouldn't even have bothered with `playerId` if I had written cache today, fwiw.

Issue #496, opened 2024-02-10
[Suggestion]  add cooldown code


bicatcho11:
**Is your feature request related to a problem? Please describe.**

I think it would be a useful addition cooldown code
The code can be used in more than one place
In thefts, for example

**Describe the solution you'd like**

**Additional context**

PR #501, opened 2024-02-14
added cooldown feature

Mirrrrrow:
Saw the suggestion in #496 and also needed this.
Setting Cooldowns: `lib.cooldown.carry = 3000`. `print(lib.cooldown.carry) --true or false depending if the cooldown is active`


Moozdzn:
This can already be achieved with custom cache values

```lua
cache('carry', function() return true end, 3000)
print(cache.carry)
```

Mirrrrrow:
yeah true mb

PR #502, opened 2024-02-15
fix(imports/require): return `false` when the code chunk returns `false`

D4isDAVID:
`true` should be the fallback value when the code chunk returns `nil`, but that should not be the case when it returns `false`.


FjamZoo (approved):

FjamZoo:
@thelindat This is a real og pr

PR #509, opened 2024-03-02
feat(interface): add locales for progress and radial menu

antond15:
Added locale entries for the default keybind labels.


LukeWasTakenn (changes requested):
Can you add the english entry to all the locale files to avoid weird behavior?

PR #512, opened 2024-03-08
fix bug with wrong detection keys on cyrillic

qwezertino:
This simple solution for detecting a cyrillic keys and convert it to right way

### DIFF
--- a/web/src/features/skillcheck/indicator.tsx
+++ b/web/src/features/skillcheck/indicator.tsx
@@ -21,10 +21,21 @@ const Indicator: React.FC<Props> = ({ angle, offset, multiplier, handleComplete,
       }),
     1
   );
-
   const keyHandler = useCallback(
     (e: KeyboardEvent) => {
-      setKeyPressed(e.key.toLowerCase());
+      const capitalHetaCode = 880;
+      const isNonLatin = e.key.charCodeAt(0) >= capitalHetaCode;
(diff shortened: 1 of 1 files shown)


web/src/features/skillcheck/indicator.tsx
```diff
@@ -21,10 +21,21 @@ const Indicator: React.FC<Props> = ({ angle, offset, multiplier, handleComplete,
       }),
     1
   );
-
   const keyHandler = useCallback(
     (e: KeyboardEvent) => {
-      setKeyPressed(e.key.toLowerCase());
+      const capitalHetaCode = 880;
```
LukeWasTakenn, line 26: Not sure if I'm missing something but that's the reasoning behind this seemingly random number, why 880?

qwezertino: Coz of unicode for symbols, all cyrillic is >= 880
Exmple from test code - 

KeyboardEvent: key='e' | code='KeyE' | charCodeAt(0) ='101'

KeyboardEvent: key='у' | code='KeyE' | charCodeAt(0) ='1091' -- there is cyrillic letter 'У' its in the same place on keyboard where eng 'E'

LukeWasTakenn: Right, makes sense.

PR #513, opened 2024-03-08
Added searchable prop to select and multi-select input types

Artichy:
As title mentions, this PR just adds the possibility to set the searchable prop to the 'select' and 'multi-select' input dialogs.
It's just an already existing feature in Mantine that is very helpful when the options list is too large.


LukeWasTakenn (approved):
Looks good, would you mind making a PR to the docs as well?

Artichy:
> Looks good, would you mind making a PR to the docs as well?

Created PR to the docs as requested:
https://github.com/overextended/overextended.github.io/pull/174

PR #515, opened 2024-03-11
fix(progress/client): sync props with statebags

Andyyy7666:
Changed props spawning from networked on client to just spawn non networked on clients using statebag

### DIFF
--- a/resource/interface/client/progress.lua
+++ b/resource/interface/client/progress.lua
@@ -206,6 +206,8 @@ end)
 RegisterKeyMapping('cancelprogress', 'Cancel current progress bar', 'keyboard', 'x')
 
 AddStateBagChangeHandler("lib:progressProps", nil, function(bagName, key, value, reserved, replicated)
+    if replicated then return end
+
     local ply = GetPlayerFromStateBagName(bagName)
     if ply == 0 then return end
 
--- a/resource/interface/client/progress.lua
+++ b/resource/interface/client/progress.lua
@@ -2,6 +2,7 @@ local progress
 local DisableControlAction = DisableControlAction
 local DisablePlayerFiring = DisablePlayerFiring
 local playerState = LocalPlayer.state
+local createdProps = {}
 
 ---@class ProgressPropProps
 ---@field model string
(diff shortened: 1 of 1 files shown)


Andyyy7666:
@thelindat tested with multiple players, no issues found.

resource/interface/client/progress.lua
```diff
@@ -131,14 +122,7 @@ local function startProgress(data)
...
     if data.prop then
-        local n = #data.prop
-        for i = 1, n > 0 and n or 1 do
-            local prop = data['prop'..i]
-
-            if prop then
-                DeleteEntity(prop)
-            end
-        end
+        playerState:set("lib:progressProps", nil, true)
```
thelindat, line 125: I use `'` for strings, not `"`.

resource/interface/client/progress.lua
```diff
@@ -220,3 +204,49 @@ RegisterCommand('cancelprogress', function()
...
+            DeleteEntity(prop)
+        end
+    end
+end
+
+RegisterNetEvent('onPlayerDropped', function(serverId)
+    deleteProgressProps(createdProps[serverId])
+end)
+
+AddStateBagChangeHandler("lib:progressProps", nil, function(bagName, key, value, reserved, replicated)
```
thelindat, line 222: Wrong string style again.

thelindat (changes requested):

FjamZoo:
After clearing the props in `deleteProgressProps` you should probably wipe the cached data aswell, as we don't want to store props that no longer exsist

Andyyy7666:
@thelindat done!

resource/interface/client/progress.lua
thelindat: > "lib:progressProps"
Not resolved.

resource/interface/client/progress.lua
```diff
@@ -220,3 +204,50 @@ RegisterCommand('cancelprogress', function()
...
+
+local function deleteProgressProps(playerProps)
+    if not playerProps then return end
+    for i = 1, #playerProps do
+        local prop = playerProps[i]
+        if DoesEntityExist(prop) then
+            DeleteEntity(prop)
+        end
+    end
+    playerProps = nil
```
thelindat, line 216: This does basically nothing. You should be assigning `createdProps[id] = nil` rather than the variable referencing it.

thelindat (changes requested):

PR #529, opened 2024-03-30
feat(import/callback): add security key verify when triggering callback

RrybaN:
This change adds security measures to prevent manual triggering of callbacks from executors using events created from lib.callback.register

### DIFF
--- a/imports/callback/client.lua
+++ b/imports/callback/client.lua
@@ -1,6 +1,12 @@
 local events = {}
 local timers = {}
 local cbEvent = ('__ox_cb_%s')
+local keyEvent = ('__ox_cb_key_%s')
+local currentKey
+
+RegisterNetEvent(keyEvent:format(cache.resource), function(key)
+	currentKey = key
+end)
 
 RegisterNetEvent(cbEvent:format(cache.resource), function(key, ...)
 	local cb = events[key]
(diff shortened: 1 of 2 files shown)


thelindat:
If the security is to just add an additional event handler, what prevents them from triggering that event?

RrybaN:
Screenshot 2024-03-30 104914

On the screenshot provided above, if they trigger the event manually using an executor with this:
`TriggerServerEvent('__ox_cb_ch_policejob:SavePlaceObject')`

This will trigger the client event without key verification:
`TriggerClientEvent(cbEvent:format(resource), source, key, callbackResponse(pcall(cb, source, ...)))`

Essentially, the pcall function is used to instantly test if the callback function has an error. Therefore, it will execute the callback function from the lib.callback.register.

thelindat:
Yes but what prevents them from triggering the keyevent and the callback event?

RrybaN:
Yes, I also considered that if multiple callbacks are triggered simultaneously, it might fail. However, I tested it by implementing numerous loops that trigger the key event using math.random to imitate wrong key. Despite triggering the key event multiple times in loops, I ensured that the callback was triggered correctly. As shown in the screenshot below, the callback did not fail.

Screenshot 2024-03-30 170433
Screenshot 2024-03-30 170450

RrybaN:
> Yes but what prevents them from triggering the keyevent and the callback event? This also just fails if multiple callbacks are being triggered by several clients at once.

I have committed a fix for a possible failure with multiple callbacks. Can you review it?

PR #530, opened 2024-03-31
refactor(import/raycast): change raycast hit return from number to boolean

RrybaN:


MadsLeander:
A better way to do this would probably be to put the equals check on/in the same line as the return, rather than redefining the variable itself.

So maybe something like this would be better?
`return hit == 1, entityHit, endCoords, surfaceNormal, materialHash`

RrybaN:
yes, that's another way but i have done that because this is a public repo with a lot of developers contributing to it so clarity and predictability of code is important.

thelindat:
This breaks fxv2 oal https://github.com/citizenfx/fivem/blob/master/ext/natives/codegen_out_native_lua.lua#L69

PR #531, opened 2024-04-01
Update lib.streamingRequest

hakanesnn:


thelindat:
should in fact be `format(assetType, asset)`

Issue #358, opened 2023-06-21
[Suggestion] Timer functions


marshular:
**Is your feature request related to a problem? Please describe.**
My feature is not related to any problem, but I would like to see some type of time timer function.

**Describe the solution you'd like**
Examples: 
``lib.timer.start(name, time)``
``lib.timer.end(name)``
``lib.timer.pause(name)``
``lib.timer.change(name, time)``
``lib.timer.onend(name, function()
-- do something
end)``

**Additional context**
N/A

PR #539, opened 2024-04-07
timer module

Demigod916:
while looking throught the issues for ox_lib i saw that this module was a suggested feature #358 from a while ago. was bored so i decided to create it.

# lib.timer
 - create both async and sync timers to trigger delayed functions
 - pause and play timers after creation
 - force timers to end early and optionally trigger the onEnd cb function still
 
 this module aims reduce the amount of reused code for timers in resources as well as provide useful methods, extending the utility of the normal timers youd see:
 ```lua
 -- traditional timer in a resource
 local start = GetGameTimer()
 local timer = 1000

 while GetGameTimer() - start > timer do
     Wait(0)
 end
 
 -- do something
 ```

 
 ### lib.timer examples:
 
 ```lua
 --sync

print('before')

local timer = lib.timer(5000, function()
    print(GetGameTimer() - start)
end)

print('after')
 ```

```lua
-- async

local start = GetGameTimer()

local timer = lib.timer.async(5000, function()
    print(GetGameTimer() - start)
end)

timer:pause()

Wait(1000)

timer:play()
    
--timer finishes in 6 seconds rather than 5 because of the pause
```

```lua
RegisterCommand('testAllTimerFeatures', function(source, args, raw)
    -- Create and start a synchronous timer
    print('Sync timer started with 3 seconds')
    local syncTimer = lib.timer(3000, function()
        print('Sync timer on end')
    end)

    -- Create and start an asynchronous timer
    local start = GetGameTimer()
    local asyncTimer = lib.timer.async(5000, function()
        print('Async timer on end', GetGameTimer() - start)
    end)
    print('Async timer started with 5 seconds')

    Wait(1000)
    print('1 second elapsed')
    print('Is the async timer paused?', asyncTimer:isPaused(), asyncTimer:getTimeLeft('ms'))


    -- Check time left in different formats for the async timer
    print('Async timer time left (ms):', asyncTimer:getTimeLeft('ms'))
    print('Async timer time left (s):', asyncTimer:getTimeLeft('s'))

    -- Pause the asynchronous timer
    asyncTimer:pause()
    print('Async timer paused')

    Wait(2000)
    print('2 more seconds elapsed')

    -- Resume the asynchronous timer
    asyncTimer:play()
    print('Async timer resumed')
    print('Async timer time left (ms):', asyncTimer:getTimeLeft('ms'))
    print('Async timer time left (s):', asyncTimer:getTimeLeft('s'))

    Wait(1000)
    print('1 second after resume')

    -- Restart the asynchronous timer
    asyncTimer:restart()
    print('Async timer restarted')

    Wait(1000)
    print('1 second after restart')

    -- Print the time left in all formats for the asynchronous timer
    local timeLeft = asyncTimer:getTimeLeft()
    print(string.format('Time left - ms: %.2f, s: %.2f, m: %.2f, h: %.2f', timeLeft.ms, timeLeft.s, timeLeft.m, timeLeft.h))

    -- Finally force end the asynchronous timer without triggering the onEnd function
    asyncTimer:forceEnd(true)
    print('Async timer force ended with onEnd')
end)

```

will write the docs for it as well if it is approved.

### DIFF
--- a/imports/timer/shared.lua
+++ b/imports/timer/shared.lua
@@ -1,17 +1,31 @@
+---@class TimerPrivateProps
+---@field onEnd? fun() cb function triggered when the timer finishes
+---@field async? boolean wether the timer should run asynchronously or not
+---@field startTime number the gametimer stamp of when the timer starts. changes when paused and played
+---@field triggerOnEnd boolean set in the forceEnd method using the optional param. wether or not the onEnd function is triggered when force ending the timer early
+---@field currentTime number current timer length
+---@field tickActive boolean or not the tick is running
+
 ---@class CTimer : OxClass
+---@field private TimerPrivateProps
+---@field constructor fun(self: self, time: number, onEnd: fun(), async: boolean)
+---@field start fun(self: self) starts the timer
+---@field tick fun(self: self) handles the tick of the timer. not for use externally
+---@field forceEnd fun(self: self, triggerOnEnd: boolean) end timer early and optionally trigger the onEnd function still
+---@field pause fun(self: self) pauses the timer until play method is called
+---@field play fun(self: self) resumes the timer if paused
+---@field getTimeLeft fun(self: self): number returns the time left on the timer in milliseconds
 local timer = lib.class('CTimer')
 
----@param time number
----@param onEnd? fun(data: CTimer)
----@param async? boolean
 function timer:constructor(time, onEnd, async)
-    if not time then
-        return lib.print.error('No time was provided to the timer')
-    end
+    assert(type(time) == "number" and time > 0, "Time must be a positive number")
+    assert(onEnd == nil or type(onEnd) == "function", "onEnd must be a function or nil")
+    assert(type(async) == "boolean" or async == nil, "async must be a boolean or nil")
 
     self.time = time
     self.paused = false
 
+    self.private.currentTime = time
     self.private.onEnd = onEnd
     self.private.async = async
     self.private.startTime = 0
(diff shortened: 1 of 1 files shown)


imports/timer/shared.lua
```diff
@@ -0,0 +1,118 @@
+---@class TimerPrivateProps
+---@field onEnd? fun() cb function triggered when the timer finishes
+---@field async? boolean wether the timer should run asynchronously or not
+---@field startTime number the gametimer stamp of when the timer starts. changes when paused and played
+---@field triggerOnEnd boolean set in the forceEnd method using the optional param. wether or not the onEnd function is triggered when force ending the timer early
+---@field currentTime number current timer length
+---@field tickActive boolean or not the tick is running
+
+---@class CTimer : OxClass
```
thelindat, line 9: Since LLS classes are applied globally I try to avoid potential conflicts by using Ox for classes (i.e. OxTimer).

imports/timer/shared.lua
```diff
@@ -0,0 +1,118 @@
...
+    if self.private.async then
+        CreateThread(function()
+            self:tick()
+        end)
+    else
+        self:tick()
+    end
+end
+
+function timer:tick()
```
thelindat, line 51: Just make this a local function that takes a timer, rather than putting it into the class prototype.

thelindat (changes requested):

imports/timer/shared.lua
```diff
@@ -0,0 +1,117 @@
...
+end
+
+---@param timerInstance OxTimer
+local function tick(timerInstance)
+    while true do
+        if timerInstance.paused then
+            while timerInstance.paused do
+                Wait(0)
+            end
+        end
```
thelindat, line 43: No need for the if check since while will satisfy the condition.

imports/timer/shared.lua
```diff
@@ -0,0 +1,117 @@
...
+            break
+        end
+
+        Wait(0)
+    end
+
+    timerInstance:onEnd()
+end
+
+function timer:start()
```
thelindat, line 55: Perhaps `start` should just take an async argument to run in a new thread?

imports/timer/shared.lua
thelindat: also what happens if you pause the timer but also force it to end?

thelindat (changes requested):

Demigod916:
it seems i didnt actually push the last update i did. let me do that now

imports/timer/shared.lua
Demigod916: > also what happens if you pause the timer but also force it to end?

latest update will unpause and set the startTime to 0

thelindat:
May make some minor changes after I play around with it but seems good for now.
Could you submit a PR for the docs?

Demigod916:
> Could you submit a PR for the docs?

yeah ill get some docs for ir

PR #542, opened 2024-04-08
feat(notify): add option to show duration

RrybaN:
**lib.notify New Option**
```typescript
showDuration?: boolean -- Option to show duration bar
```

**Example:**
```lua
lib.notify({
    title = 'Custom',
    description = 'Custom',
    showDuration = true,
    icon = 'eye',
    iconColor = '#C53030'
})
```

**Preview:**

Screenshot 2024-04-08 231022

### DIFF
--- a/web/src/features/notifications/NotificationWrapper.tsx
+++ b/web/src/features/notifications/NotificationWrapper.tsx
@@ -1,7 +1,7 @@
 import { useNuiEvent } from '../../hooks/useNuiEvent';
 import { toast, Toaster } from 'react-hot-toast';
 import ReactMarkdown from 'react-markdown';
-import { Avatar, Box, createStyles, Group, keyframes, Stack, Text } from '@mantine/core';
+import { Avatar, Box, Center, createStyles, Group, keyframes, RingProgress, Stack, Text } from '@mantine/core';
 import React from 'react';
 import type { NotificationProps } from '../../typings';
 import MarkdownComponents from '../../config/MarkdownComponents';
@@ -103,11 +103,20 @@ const exitAnimationBottom = keyframes({
   },
 });
 
+const durationCircle = keyframes({
+  '0%': { strokeDasharray: `0, ${15.1 * 2 * Math.PI}` },
+  '100%': { strokeDasharray: `${15.1 * 2 * Math.PI}, 0` },
+});
+
 const Notifications: React.FC = () => {
   const { classes } = useStyles();
 
   useNuiEvent<NotificationProps>('notify', (data) => {
     if (!data.title && !data.description) return;
+
+    let iconColor: string;
+    let duration = data.duration || 3000;
+
     // Backwards compat with old notifications
     let position = data.position;
     switch (position) {
@@ -134,6 +143,24 @@ const Notifications: React.FC = () => {
           break;
       }
     }
+    if (!data.iconColor) {
+      switch (data.type) {
+        case 'error':
+          iconColor = 'red';
+          break;
+        case 'success':
+          iconColor = 'teal';
+          break;
+        case 'warning':
+          iconColor = 'yellow';
+          break;
+        default:
+          iconColor = 'blue';
+          break;
+      }
+    } else {
+      iconColor = data.iconColor;
+    }
     toast.custom(
       (t) => (
         <Box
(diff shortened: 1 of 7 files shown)


Demigod916:
hot

FjamZoo:
Good looks

LukeWasTakenn (commented):
Code looks good, though there is this interesting issue of not using a mantine color so if you  use a custom hex color and want to display the notification progress it just looks weird.

image

```lua
  debugData<NotificationProps>([
    {
      action: 'notify',
      data: {
        title: 'Success',
        description: 'Notification description',
        type: 'success',
        id: 'pogchamp',
        duration: 200000,
        showDuration: true,
        iconColor: '#d5eb31',
      },
    },
  ]);
  debugData<NotificationProps>([
    {
      action: 'notify',
      data: {
        title: 'Success',
        description: 'Notification description',
        type: 'success',
        duration: 200000,
        showDuration: false,
        iconColor: '#d5eb31',
      },
    },
  ]);
```

Not entirely sure what we can do about this, maybe perhaps what mantine is doing and just set the background color of the avatar to the same color and lover the alpha?

I'm open to suggestions.

RrybaN:
I think the problem here is that the mantine version is not on the latest version, the latest version can use rgba or hex on `Avatar` component. But either way i found a another solution for this, instead of using `Avatar` component i used `ThemeIcon` for this.

**Preview:**

Screenshot 2024-04-11 191246

web/src/features/notifications/NotificationWrapper.tsx
```diff
@@ -103,11 +103,24 @@ const exitAnimationBottom = keyframes({
...
 });
 
+const durationCircle = keyframes({
+  '0%': { strokeDasharray: `0, ${15.1 * 2 * Math.PI}` },
+  '100%': { strokeDasharray: `${15.1 * 2 * Math.PI}, 0` },
+});
+
+const isValidHexColor = (hex: string) => {
+  return /^#[0-9A-F]{6}$/i.test(hex)
+}
```
LukeWasTakenn, line 113: Hex or mantine color aren't the only options, you can pass colors as rgb or hsl like such: `rgb(255, 255, 110)` which results in the icon not being colored

image

LukeWasTakenn (changes requested):
Looks great overall though I can also see someone not wanting to show the progress but wanting the light background color of ThemeIcon and I'm not sure what would be the best way to handle that 🤔 

Since the size of the icon is now a bit different it looks weird if you have a notification with an icon background and another one without it.

image


Maybe just constantly show the background color with ThemeIcon? Though I can also see some people *not* preferring this.

RrybaN:
Noted, but which one do you prefer the light background or constantly show the background color? Or do you have anything else in your mind.

web/src/features/notifications/NotificationWrapper.tsx
RrybaN: Oh yes, i clearly forgot about this

LukeWasTakenn:
> Noted, but which one do you prefer the light background or constantly show the background color? Or do you have anything else in your mind.

The light background color is good the way it is now, just have it always present even when using `iconColor` that way we avoid the difference in spacing, and also need to make it work with different color formats.

I think we also make it display duration by default and make it an optional opt out by setting `showDuration: false`.

RrybaN:
> and also need to make it work with different color formats.

For this one can i add a package `tinycolor2`
https://www.npmjs.com/package/tinycolor2

so we can just convert all available color formats to rgba, because the current mantine version only accepts hex, rgb, rgba

RrybaN:
For this one show duration is defaulted to show and also background color is always present even when using `iconColor`. Can now use `'hex' | 'hexa' | 'rgb' | 'rgba' | 'hsl' | 'hsla'` for `iconColor`.

**Preview:**

Screenshot 2024-04-12 040806

thelindat (approved):
LGTM, just need to add showDuration to the Lua and TS NotifyProps type.

LukeWasTakenn (approved):
Looks great, thanks for the PR!

Would you also mind updating the docs with the `showDuration` prop?

RrybaN:
Sure, I don't mind updating the docs. I'm glad to help with anything.

PR #555, opened 2024-04-23
chore(locale): update no.json

MadsLeander:
The `settings` locale is for some reason not included in most (if not all) non-english locale files. I'm not sure if that is intentional or not, so please let me know if you want that removed again.

The reason for changing `cancel_progress` is that "progresjonsbar" seems to be the industry standard translation for "progress bar", it also sounds a lot less clunky than "fremdriftsindikatoren".

### DIFF
--- a/locales/no.json
+++ b/locales/no.json
@@ -1,29 +1,30 @@
 {
   "language": "Norsk",
+  "settings": "Innstillinger",
   "ui": {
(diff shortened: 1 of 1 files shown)


thelindat:
> not included in most (if not all) non-english locale files

New locale, no translations. Previously I was adding the English string as a placeholder, but that's no longer necessary.

PR #558, opened 2024-04-24
Add onSelect function for dialog select rows

Scullyy:
This will allow you to trigger a function when an option is selected, example usage:

```lua
lib.inputDialog('Feature Example', {
    {
        type = 'select',
        required = true,
        icon = 'face-smile',
        label = 'Mood',
        default = 'angry',
        placeholder = 'Select',
        options = {
            { label = 'Angry', value = 'angry' },
            { label = 'Happy', value = 'happy' }
        },
        onSelect = function(value)
            print(('Linden is %s today!'):format(value)) -- Linden is angry | happy today!
        end
    }
})
```

### DIFF
--- a/resource/interface/client/input.lua
+++ b/resource/interface/client/input.lua
@@ -1,9 +1,11 @@
 local input
+local currentRows
 
 ---@class InputDialogRowProps
 ---@field type 'input' | 'number' | 'checkbox' | 'select' | 'slider' | 'multi-select' | 'date' | 'date-range' | 'time' | 'textarea' | 'color'
 ---@field label string
 ---@field options? { value: string, label: string, default?: string }[]
+---@field onSelect? fun(args: any)
 ---@field password? boolean
 ---@field icon? string | {[1]: IconProp, [2]: string};
 ---@field iconColor? string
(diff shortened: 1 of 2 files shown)


LukeWasTakenn:
Input dialog is supposed to be a form where you do your own actions when submitted, this is not something we want.

Scullyy:
> Input dialog is supposed to be a form where you do your own actions when submitted, this is not something we want.

This is very useful, for example:

https://i.gyazo.com/5ff957c8d7acd6f64d0c00508b736303.gif

thelindat:
For anybody coming across this wondering why we don't want this:
- restricting to just the "select" field type is too limiting
- making the function available to other field types adds complexity
- most importantly, this will get misused and inevitably lead to us having to handle nonsense issues from third party resources

Such features need to be properly discussed, thought out, planned, and implemented. The given use-case is very simple and niche, but also not enough of a reason for us to implement something half-assed.

The user's response elsewhere after this was closed also killed any chance of talking such a feature out.

Scullyy:
- I would have been happy to make this compatible with all select types if this was communicated instead of just immediately closed, I’m always open to hearing improvements that I can make.
- I feel like anything can be abused if used the wrong way but I would have been happy to discuss any methods that could have been used to circumvent this if the topic was open to conversation. 

I wouldn’t consider it half-assed but I see your point. I feel there was a better approach to this on both sides, with this being immediately closed with no room for conversation and my response (which I apologise for, it has been a long day) it will be troublesome to try to converse the points you mentioned to get this into a considerable state.

Issue #554, opened 2024-04-22
[Bug] Notification progress circle timer not reset.


zf-development:
**Describe the bug**
The circle around the icon in the notifications is nice but, when multiple notifications are sent with the same id, the timer is not reset so it looks like this even if the notification is going to be there for 10 more seconds for exemple
![image](https://github.com/overextended/ox_lib/assets/48424078/3ba170b0-298d-474c-9c97-63956c8b34d7)

The circle around the background of the icon is not there anymore.

**To Reproduce**
1. Send multiple notifications with the same id.

**Expected behavior**
The timer will start for the first notification sent but it you send another one with the same id while the timer is going down, it won't be reset.

**Screenshots**
![image](https://github.com/overextended/ox_lib/assets/48424078/d8edc848-c863-4395-b33a-4062a55a054f)

zf-development:
Here's a video to show it live.
Bottom left, you can see when I press E to send a notification.
Top Right, you can see the timer and notification.
https://github.com/overextended/ox_lib/assets/48424078/9c8e22dc-7c75-4c20-ba13-3fc71946f967

LukeWasTakenn:
@RrybaN You wanna have a look at this one?

RrybaN:
Okay, will have this fixed.

PR #559, opened 2024-04-26
fix(notify): reset circle animation on notification re-trigger with id

RrybaN:
This pull request is a fix for issue #554 

### **Fixed**
Reset circle animation when re-triggering a notification with `ID`.

### **Solution**
Assigning unique keys to each `RingProgress` instance allows React to properly manage its state and trigger re-renders when necessary.

### **Preview**
notify


LukeWasTakenn (approved):
Looks good, I've pushed a commit that changes the key counter to be a ref instead of state to avoid unnecessary rerenders.

PR #567, opened 2024-05-01
Update shared.lua

l-404-l:
Fix restart not keeping async state


thelindat:
You could just make restart take a bool for running async? You may want to be able to swap between sync and async for whatever reason.

PR #570, opened 2024-05-04
linear interpolation for vectors.

Demigod916:
`math.lerpVectors` is a standard linear interpolation between 2 vectors. This will allow developers to smoothy transition between 2 vectors for rotations or coords over a set duration.

this function is currently blocking. optional async is something that we can add if desired.

example: 
```lua
RegisterCommand('testLerp', function(source, args, raw)
	lib.requestModel(`gr_prop_gr_target_04b`)

	local offset = GetOffsetFromEntityInWorldCoords(cache.ped, 0, 3.0, 0)
	local object = CreateObject(`gr_prop_gr_target_04b`, offset.x, offset.y, offset.z, false, false, false)

	PlaceObjectOnGroundProperly_2(object)
	math.lerpVectors(vec3(0, 0, 0), vec3(-90.0, 0, 0), 500, function(interpolated)
		SetEntityRotation(object, interpolated.x, interpolated.y, interpolated.z, 0, false)
	end)
end)
```

https://github.com/overextended/ox_lib/assets/60995067/3d1d06a7-4458-4f35-8519-be37a3a5f6f8

### DIFF
--- a/imports/math/shared.lua
+++ b/imports/math/shared.lua
@@ -134,4 +134,22 @@ function math.clamp(val, lower, upper) -- credit https://love2d.org/forums/viewt
     return math.max(lower, math.min(upper, val))
 end
 
+---Lerps (linearly interpolates) between two vectors over a specified duration.
+---@param startVector vector -- Starting vector, the interpolation begins here.
+---@param endVector vector   -- Ending vector, the interpolation targets this.
+---@param duration number    -- The total duration of the interpolation in milliseconds.
+---@param cb function        -- Callback function that gets called with the interpolated vector as its argument.
+---@return boolean
+function math.lerpVectors(startVector, endVector, duration, cb)
+	local startTime = GetGameTimer()
+	local t = 0
+	while t < 1 do
+		t = math.min((GetGameTimer() - startTime) / duration, 1)
+		local interpolatedVector = startVector + (endVector - startVector) * t
+		cb(interpolatedVector)
+		Wait(0)
+	end
+	return true
+end
+
 return lib.math
--- a/imports/math/shared.lua
+++ b/imports/math/shared.lua
@@ -120,8 +120,8 @@ end
 ---@param seperator? string
 ---@return string
 function math.groupdigits(number, seperator) -- credit http://richard.warburton.it
-    local left,num,right = string.match(number,'^([^%d]*%d)(%d*)(.-)$')
-    return left..(num:reverse():gsub('(%d%d%d)','%1' .. (seperator or ',')):reverse())..right
+    local left, num, right = string.match(number, '^([^%d]*%d)(%d*)(.-)$')
+    return left .. (num:reverse():gsub('(%d%d%d)', '%1' .. (seperator or ',')):reverse()) .. right
 end
 
 ---Clamp a number between 2 other numbers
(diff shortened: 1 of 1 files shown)


thelindat:
An iterator might look a bit nicer in this situation.

```lua
for interpolated in math.lerpVectors(vec3(0, 0, 0), vec3(-90.0, 0, 0), 500) do
    SetEntityRotation(object, interpolated.x, interpolated.y, interpolated.z, 0, false)
end)
```

I would also prefer `math.lerp` that can handle interpolation of scalars, vectors, or even tables (both arrays and hashmaps, to a limit).

Demigod916:
so for the iterator method, it would use a coroutine?

thelindat:
You basically just return a function instead of having a function argument; it calls the function until it gets a nil value.

Demigod916:
okay updated.

```lua
-- Example usage for a scalar
for value in math.lerp(0, 100, 500) do
    print(value)
end

-- Example usage for a vector (assuming vector operations are defined)
for vec in math.lerp(vec3(0, 0, 0), vec3(10, 10, 10), 1000) do
    print(vec.x, vec.y, vec.z)
end

-- Example usage for a table
for tbl in math.lerp({x=0, y=0}, {x=10, y=20}, 1000) do
    print(tbl.x, tbl.y)
end
```

Demigod916:
okay updated.

```lua
for value in math.lerp(0, 100, 500) do
    print(value)
end

for vec in math.lerp(vec3(0, 0, 0), vec3(10, 10, 10), 1000) do
    print(vec.x, vec.y, vec.z)
end

for tbl in math.lerp({x=0, y=0}, {x=10, y=20}, 1000) do
    print(tbl.x, tbl.y)
end
```

imports/math/shared.lua
```diff
@@ -134,4 +134,46 @@ function math.clamp(val, lower, upper) -- credit https://love2d.org/forums/viewt
...
+			return s + (f - s) * t
+		elseif type(s) == "table" then
+			local result = {}
+			for i in pairs(s) do
+				result[i] = s[i] + (f[i] - s[i]) * t
+			end
+			return result
+		else
+			lib.print.error("unsupported type for lerp: " .. type(s))
+		end
```
thelindat, line 161: Can just end the `if` statement then trigger the error. Should just use a regular `error` rather than `lib.print` which is more for simple console output and doesn't have stack traces and whatnot.

I also have a typeError function defined somewhere you could copy over or expose, as this error message leaves much to be desired.

imports/math/shared.lua
```diff
@@ -134,4 +134,46 @@ function math.clamp(val, lower, upper) -- credit https://love2d.org/forums/viewt
...
+---@param start number|table|vector2|vector3|vector4  -- The starting value of the interpolation.
+---@param finish number|table|vector2|vector3|vector4 -- The ending value of the interpolation.
+---@param duration number                             -- The duration over which to interpolate.
+---@return function                                   -- a function that returns interpolated values over time.
+function math.lerp(start, finish, duration)
+	local startTime = GetGameTimer()
+	local typeStart = type(start)
+
+	if typeStart ~= type(finish) then
+		lib.print.error("start and finish must be of the same type")
```
thelindat, line 147: Should probably be more descriptive, and use the `error` function for full stack-trace and error handling. See some other error messages I've written for inspiration.

imports/math/shared.lua
thelindat, line 161: Not sure about using a locally scoped function here.

imports/math/shared.lua
```diff
@@ -134,4 +134,46 @@ function math.clamp(val, lower, upper) -- credit https://love2d.org/forums/viewt
...
+	return function()
+		if last then return nil end
+		Wait(0)
+		local t = (GetGameTimer() - startTime) / duration
+		t = math.min(t, 1)
+		if t == 1 then
+			last = true
+			return finish
+		end
+		return interpolate(start, finish, t)
```
thelindat, line 174: This is going to type-check on each frame which, while minimal, seems pretty unnecessary.

imports/math/shared.lua
thelindat: See the type-checking note below from calling this function. Probably better to define a couple of functions in the main scope and then just assign that function to a local variable depending on the type.

Side note you have `typeStart` defined above and could reuse that.

thelindat (changes requested):
I don't mind making these changes myself, but here's some nitpicks.

PR #588, opened 2024-05-25
fix(imports/logger): Loki encode does not work as expected

Cata-a1138:
Solve the problem that tags are converted to “table:xxx” by string.tostringall and therefore can't be encoded to json by encode.
test code
```lua
lib.logger(-1, 'LogTest', 'Test String', 'test123:test123', 'rpname:Lena Meier', 'job:police')
lib.logger(-1, 'LogTest', 'Test Table', { rpname = "Lena Meier", job = "police" })
``` 
result
图片

### DIFF
--- a/imports/logger/server.lua
+++ b/imports/logger/server.lua
@@ -246,8 +246,13 @@ if service == 'loki' then
         local values = {}
 
         -- Format the args into strings
-        local tags = formatTags(source, ... and string.strjoin(',', string.tostringall(...)) or nil)
-        local tagsTable = convertDDTagsToKVP(tags)
+        local tagsTable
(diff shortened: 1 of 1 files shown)


SeaLife:
Did you upgrade to Loki 3.0.0 or is this working on 2.8.x now as well?
If not, we should probably mention this in the documentation or provide a additional loki2.8.x provider or smth. like that :)
(Which is not in scope of this change i guess, just something we should consider)

thelindat:
I've reverted the last change to loki logging due to backwards compatibility issues, and closing this just to avoid issues.
Logging is long overdue for a rewrite and it will happen sometime after the nodejs 20 update.

PR #597, opened 2024-06-18
feat(addCommand): longString

BerkieBb:
Whenever you have a command that has a text type at the end where you want to catch all text from that argument and on, only the first word will be assigned to that argument and the rest ends up all in an array which makes for a mixed table. With longString it catches those extras if the argument is at the end of the given params.

For example
`/test test1 test2`

Before only `test1` would be shown.
After, all text will be shown.

*I haven't tested this yet but ChatGPT said it will work fine*

### DIFF
--- a/imports/addCommand/server.lua
+++ b/imports/addCommand/server.lua
@@ -46,14 +46,12 @@ local function parseArguments(source, args, raw, params)
                 value = false
             end
         elseif param.type == 'longString' and i == paramsNum then
-            value = arg
-
-            local argsNum = #args
-            if paramsNum < argsNum then
-                for a = i + 1, argsNum do
-                    value = value .. ' ' .. args[a]
-                end
+            local words = {}
+            for word in raw:gmatch('%S+') do
+                words[#words + 1] = word
             end
+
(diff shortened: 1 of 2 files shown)


imports/addCommand/server.lua
```diff
@@ -44,6 +45,15 @@ local function parseArguments(source, args, raw, params)
...
             end
+        elseif param.type == 'longString' and i == paramsNum then
+            value = arg
+
+            local argsNum = #args
+            if paramsNum < argsNum then
+                for a = i + 1, argsNum do
+                    value = value .. ' ' .. args[a]
+                end
+            end
```
thelindat, line 56: Rather than concatenating the rest of the passed arguments, just take the raw string and trim off the other arguments.

BerkieBb: Ah yeah I forgot we had a raw string

BerkieBb:
Raw string now used

thelindat:
To clarify, there's no need to split the string into many parts and concatenate or join strings together. Just get the position of the last argument in the raw string and then make a substring from that point.

BerkieBb:
Changed

PR #603, opened 2024-06-23
feat(math): Add math.round function

jag3dagster:
- Handles whole numbers and decimal places


jag3dagster:
Matching docs PR: https://github.com/overextended/overextended.github.io/pull/197

PR #604, opened 2024-06-25
(Feat)Add table.length function

Luke20201:
LUA doesn't have this, and I thought it'd be nice for a FiveM library to have it so I committed to ox.


thelindat:
Well `#tbl` for actual length (assuming an continuous array without gaps). For hashmaps I'd rather a specialised class (like array) with methods for such things.

PR #616, opened 2024-07-07
fix(addCommand): print caught tables instead of table references

Manason:
When the error thrown by the command handler function is a table, this is printed by reference. This solves the problem json encoding the error body.

### DIFF
--- a/imports/addCommand/server.lua
+++ b/imports/addCommand/server.lua
@@ -113,7 +113,7 @@ function lib.addCommand(commandName, properties, cb, ...)
         local success, resp = pcall(cb, source, args, raw)
 
         if not success then
-            Citizen.Trace(("^1command '%s' failed to execute!\n%s"):format(string.strsplit(' ', raw) or raw, resp))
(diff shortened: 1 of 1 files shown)


thelindat:
... why are your errors tables?

Manason:
> ... why are your errors tables?

We have error code + error message

Manason:
> Although you can use a value of any type as an error message, usually error messages are strings describing what went wrong
https://www.lua.org/pil/8.5.html

thelindat:
> usually error messages are string

Now go send a table to the error function and see what fxserver does with it.

thelindat:
```lua
CreateThread(function()
    error({ 'this must be a bug bro' })
end)

---@class Error : OxClass
local Error = lib.class('Error')

function Error:constructor(obj)
    for k, v in pairs(obj) do
        self[k] = v
    end
end

function Error:__tostring()
    return json.encode(self, { indent = true, sort_keys = true })
end

function Error:throw()
    return error(('Error: %s'):format(self), 2)
end

CreateThread(function()
    Error:new({ message = 'work smarter', code = 1 }):throw()
end)
```

PR #621, opened 2024-07-16
Update server.lua to add fivemerr support for logs

ItsTrapson:


thelindat:
I won't be merging fivemerr support. I'm not going to get into the reasons here.
They are, of course, free to provide their own script file and instructions for setup.

PR #623, opened 2024-07-19
Added convar for radial default keybind.

GttMone:


Moozdzn:
Discussed in https://github.com/overextended/ox_lib/pull/257

PR #624, opened 2024-07-29
feat(resource/vehicleProperties/client.lua): adding a configuration for plate format

HR0110:
…ace at the end of the plate value that breaks the value matchings

### DIFF
--- a/resource/vehicleProperties/client.lua
+++ b/resource/vehicleProperties/client.lua
@@ -198,7 +198,7 @@ function lib.getVehicleProperties(vehicle)
 
         return {
             model = GetEntityModel(vehicle),
-            plate = GetVehicleNumberPlateText(vehicle):sub(1, GetConvarInt('ox:plateFormat', 8)),
+            plate = GetVehicleNumberPlateText(vehicle):sub(1, #GetConvar('ox:plateFormat', '........')),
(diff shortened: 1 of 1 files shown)


CeebDev:
If you use plate format with 7 digits instead 8 it's not the case for all users.

ox_lib just need to handle plate with what GTA return

HR0110:
> If you use plate format with 7 digits instead 8 it's not the case for all users.
> 
> ox_lib just need to handle plate with what GTA return

Yes I know, just forgotten to commit the part with a configuration for a 7 bits plates or 8 bits because the 7 bits format is very popular and makes many issues when trying to match with a plate text. I'm gonna do it. Thank you!

thelindat:
This is stupid and so incredibly breaking.

> the 7 bits format is very popular

That doesn't change the fact that GTAV vehicle plates are _always_ 8 characters long. Either trim the whitespace in the plate, or pad the comparison. _In your own code._

HR0110:
> This is stupid and so incredibly breaking.
> 
> > the 7 bits format is very popular
> 
> That doesn't change the fact that GTAV vehicle plates are _always_ 8 characters long. Either trim the whitespace in the plate, or pad the comparison. _In your own code._

The idea is that ox_lib can cut the plates according to the user's preference directly from it, rather than having to use ox_core for this purpose

PR #626, opened 2024-07-30
feat(resource/vehicleProperties/client.lua): Ensures vehicle property is set properly

mpcmarques:
I know that applying properties to server side vehicles can be buggy. It seems that with the latest version of this script, sometimes when changing buckets with more than 100 players, or spawning the vehicle with a person inside, it gets messy and there is a chance that the vehicle properties aren`t applied. This ensures that the correct properties are set. Tried in a server with 200 online.

Probably doesn`t fix when the vehicle is spawned with a player driving it. On my server it still needed a client to be event sent to the player to ensure all mods are applied correct.

### DIFF
--- a/resource/vehicleProperties/client.lua
+++ b/resource/vehicleProperties/client.lua
@@ -113,14 +113,28 @@ AddStateBagChangeHandler('ox_lib:setVehicleProperties', '', function(bagName, _,
     if not entityExists then return end
 
     lib.setVehicleProperties(entity, value)
-    Wait(200)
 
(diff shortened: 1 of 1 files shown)


thelindat:
We've already tried loops similar to above to no improved effect - also some weirdness to how this is written.

There's no guarantee that the props are set even if you wait 100 seconds and are the entity owner for the whole period, according to tests from various other people.

I'll be refactoring this ~soon, probably just use regular events targeting the entity owner, and some other bullshit that shouldn't be necessary but FiveM:tm:.

PR #634, opened 2024-08-14
fix color problem

iamilia:
I added another check to solve this problem completely


thelindat:
Got any more information beyond "this problem"?

iamilia:
When you use the SetVehicleColours function, it makes sense for the GetIsVehiclePrimaryColourCustom function to return false, but that's not the case. However, based on my observations, when there is a PrimaryColourCustom, the GetVehicleColours function returns the colorPrimary as 0. If colorPrimary is not zero, it means that the SetVehicleColours function has been used, and the issue with the colors is resolved.

iamilia:
And it doesn't matter whether the ClearVehicleCustomPrimaryColour function has been used or not. The same issue exists with the secondary color as well.

PR #647, opened 2024-10-01
refactor and improve performance of table functions

BigSmoKe07:
Function Refactor: 
- contains
- table_matches
- table_merge

### DIFF
--- a/imports/table/shared.lua
+++ b/imports/table/shared.lua
@@ -81,17 +81,18 @@ end
 ---@return table
 ---Merges two tables together. Defaults to adding duplicate keys together if they are numbers, otherwise they are overriden.
 local function table_merge(t1, t2, addDuplicateNumbers)
-    if addDuplicateNumbers == nil then addDuplicateNumbers = true end
-    for k, v in pairs(t2) do
-        local type1 = type(t1[k])
-        local type2 = type(v)
+    addDuplicateNumbers = addDuplicateNumbers ~= nil and addDuplicateNumbers or true
+    for k, v2 in pairs(t2) do
+        local v1 = t1[k]
+        local type1 = type(v1)
+        local type2 = type(v2)
 
 		if type1 == 'table' and type2 == 'table' then
-            table_merge(t1[k], v, addDuplicateNumbers)
+            table_merge(v1, v2, addDuplicateNumbers)
         elseif addDuplicateNumbers and (type1 == 'number' and type2 == 'number') then
-            t1[k] += v
+            t1[k] = v1 + v2
 		else
-			t1[k] = v
+			t1[k] = v2
         end
     end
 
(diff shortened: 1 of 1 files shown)


imports/table/shared.lua
```diff
@@ -9,47 +9,54 @@ local pairs = pairs
...
 ---@param t2 any
 ---@return boolean
 ---Compares if two values are equal, iterating over tables and matching both keys and values.
 local function table_matches(t1, t2)
-	local type1, type2 = type(t1), type(t2)
-
-	if type1 ~= type2 then return false end
-	if type1 ~= 'table' and type2 ~= 'table' then return t1 == t2 end
+    if type(t1) ~= type(t2) then return false end
+    if type(t1) ~= 'table' then return t1 == t2 end
```
thelindat, line 44: Considering we're micro-optimising here, there's no reason not to reuse type1.

thelindat: While we're at it we can get a small boost by comparing table types and lengths.

```lua
    local tabletype1 = table.type(t1)

    if tabletype1 ~= table.type(t2) or (tabletype1 == 'array' and #t1 ~= #t2) then
        return false
    end
```

imports/table/shared.lua
```diff
@@ -9,47 +9,57 @@ local pairs = pairs
...
 ---Compares if two values are equal, iterating over tables and matching both keys and values.
 local function table_matches(t1, t2)
-	local type1, type2 = type(t1), type(t2)
+    local tabletype1 = table.type(t1) 
 
-	if type1 ~= type2 then return false end
-	if type1 ~= 'table' and type2 ~= 'table' then return t1 == t2 end
+    if tabletype1 ~= table.type(t2) or (tabletype1 == 'array' and #t1 ~= #t2) then 
+        return false 
+    end
```
thelindat, line 47: To clarify, I meant to use the table.type check in addition to the type check.
This _should_ work as a replacement for the previous code though.

```diff
- local type1 = type(t1)

- if type1 ~= type(t2) then return false end
- if type1 ~= 'table' then return t1 == t2 end

+ local tabletype1 = table.type(t1)

+ if not tabletype1 then return t1 == t2 end

+ if tabletype1 ~= table.type(t2) or (tabletype1 == 'array' and #t1 ~= #t2) then
+     return false
+ end
```

rhykosqc:
With this update, we have encountered an inventory problem. We are no longer able to reload weapons.

--> I think its come from table.contains

```
function Inventory.GetSlotWithItem(itemName, metadata, strict)
    local inventory = PlayerData.inventory
    local item = Items(itemName) --[[@as OxClientItem?]]
    
    if not inventory or not item then return end
    
    metadata = assertMetadata(metadata)
    local tablematch = strict and table.matches or table.contains
    
    for _, slotData in pairs(inventory) do
        if slotData and slotData.name == item.name and (not metadata or tablematch(slotData.metadata, metadata)) then
	   return slotData
        end
    end
end
```

PR #650, opened 2024-10-06
feat(client/dui): implement dui handling

CeebDev:
It's really simple to create new dui and don't have to handle for resource stop etc

```lua
local dui = lib.dui:new({
	url = ("nui://%s/web/index.html"):format(cache.resource), 
	width = 1920, 
	height = 1080,
	debug = true
})

-- Change url
dui:setUrl("https://google.com")

-- Send a message
dui:sendMessage({
	action = "play",
	value = "https://www.youtube.com/watch?v=dQw4w9WgXcQ"
})

-- Destroy
dui:remove()
```

It's just a utility, all texture replace etc need to be handle by other scripts

### *JS version is not tested*

### DIFF
--- a/package/client/resource/dui/index.ts
+++ b/package/client/resource/dui/index.ts
@@ -1,82 +1,80 @@
 import {
-  GetGameTimer,
-  CreateDui,
-  GetDuiHandle,
-  CreateRuntimeTxd,
-  CreateRuntimeTextureFromDuiHandle,
-  SetDuiUrl,
-  DestroyDui,
-  SendDuiMessage,
+	GetGameTimer,
+	CreateDui,
+	GetDuiHandle,
+	CreateRuntimeTxd,
+	CreateRuntimeTextureFromDuiHandle,
+	SetDuiUrl,
+	DestroyDui,
+	SendDuiMessage,
 } from '@nativewrappers/client';
 import { cache } from '../cache';
 
 let duis: { [key: string]: Dui } = {};
+let currentId = 0;
 
 interface LibDui {
-  url: string;
-  width: number;
-  height: number;
-  debug?: boolean;
+	url: string;
+	width: number;
+	height: number;
+	debug?: boolean;
 }
 
 export class Dui {
-  id: string = "";
-  debug: boolean = false;
-  url: string = "";
-  duiObject: number = 0;
-  duiHandle: string = "";
-  runtimeTxd: number = 0;
-  txdObject: number = 0;
-  dictName: string = "";
-  txtName: string = "";
+	id: string = "";
+	debug: boolean = false;
+	url: string = "";
+	duiObject: number = 0;
+	duiHandle: string = "";
+	runtimeTxd: number = 0;
+	txdObject: number = 0;
+	dictName: string = "";
+	txtName: string = "";
 
-  constructor(data: LibDui) {
-    const time = GetGameTimer();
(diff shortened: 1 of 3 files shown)


thelindat:
Should be using ox_lib features, i.e. lib.class and cache.resource

CeebDev:
> Should be using ox_lib features, i.e. lib.class and cache.resource

I didn't know that cache has a resource prop, also, i've worked with the point file in mind which doesn't use the new lib.class, have made changes

package/client/resource/dui/index.ts
```diff
@@ -0,0 +1,80 @@
+import {
+	GetGameTimer,
+	CreateDui,
+	GetDuiHandle,
+	CreateRuntimeTxd,
+	CreateRuntimeTextureFromDuiHandle,
+	SetDuiUrl,
+	DestroyDui,
+	SendDuiMessage,
+} from '@nativewrappers/client';
```
thelindat, line 10: These types are globally defined from the `citizenfx/client` package and shouldn't be imported (and don't exist at the given path).

package/client/resource/dui/index.ts
```diff
@@ -0,0 +1,80 @@
...
+import { cache } from '../cache';
+
+let duis: { [key: string]: Dui } = {};
```
thelindat, line 13:
```ts
const duis: Record<string, Dui> = {};
```

package/client/resource/dui/index.ts
```diff
@@ -0,0 +1,80 @@
...
+let currentId = 0;
+
+interface LibDui {
```
thelindat, line 16: `DuiProperties` is better and matches your Lua type.

package/client/resource/dui/index.ts
```diff
@@ -0,0 +1,80 @@
...
+export class Dui {
+	id: string = "";
+	debug: boolean = false;
+	url: string = "";
+	duiObject: number = 0;
+	duiHandle: string = "";
+	runtimeTxd: number = 0;
+	txdObject: number = 0;
+	dictName: string = "";
+	txtName: string = "";
```
thelindat, line 32: Some of these should probably be marked as private to discourage manually setting the values. Same applies to the Lua implementation (I don't mind if it's real private or just typed as such).

package/client/resource/dui/index.ts
```diff
@@ -0,0 +1,80 @@
...
+		this.duiObject = CreateDui(data.url, data.width, data.height);
+		this.duiHandle = GetDuiHandle(this.duiObject);
+		this.runtimeTxd = CreateRuntimeTxd(this.dictName);
+		this.txdObject = CreateRuntimeTextureFromDuiHandle(this.runtimeTxd, this.txtName, this.duiHandle);
+		duis[id] = this;
+
+		if (this.debug) console.log(`Dui ${this.id} created`);
+	}
+
+	remove = () => {
```
thelindat, line 52: Proper method syntax is

```ts
remove() {

}

thelindat (changes requested):

package/client/resource/dui/index.ts
CeebDev: https://github.com/overextended/ox_lib/blob/cba4b85a2958f68a93211676eaeee88fa2d0dcd8/package/client/resource/points/index.ts#L46

PR #655, opened 2024-10-20
feat(client): lib.requestControl

someone005:
Function to request network control of an entity


thelindat:
Request natives are strongly discouraged and should not be used in any situation. Generally you should trigger an event on the entity owner, or use non-networked entities synced with events.

https://docs.fivem.net/docs/server-manual/server-commands/#sv_filterrequestcontrol-mode

Issue #651, opened 2024-10-13
[Suggestion] Add Latent Server Callback Support


FingerlessGlov3s:
**Is your feature request related to a problem? Please describe.**
You can use `TriggerLatentClientEvent` function to send data to a client that doesn't block the clients network channel. So it would be great if we could register a latent callback instead when needed. This would be more flavourful when sending large amounts of data or for non time critical events, to stop blocking of the client network channel.

**Describe the solution you'd like**
A function like `lib.callback.register` call it say `lib.callback.registerLatent`, which uses `TriggerLatentClientEvent` instead of `TriggerClientEvent` for client and `TriggerLatentServerEvent` instead of `TriggerServerEvent` for server.

Default of bps of `50000` with an argument to higher or lower it.

**Additional context**
https://docs.fivem.net/docs/scripting-manual/working-with-events/triggering-events/#triggering-latent-server-events
https://docs.fivem.net/docs/scripting-manual/working-with-events/triggering-events/#triggering-latent-client-events

Hope this makes sense 😊

PR #656, opened 2024-10-20
feat(callbacks): latent callbcks

someone005:
Add latent version of functions:
- lib.callback                       -> lib.callback.latent
- lib.callback.register          -> lib.callback.registerLatent
- lib.callback.await              -> lib.callback.latentAwait

Idea from issue [#651](https://github.com/overextended/ox_lib/issues/651)

### DIFF
--- a/imports/callback/client.lua
+++ b/imports/callback/client.lua
@@ -32,7 +32,7 @@ end
 ---@param cb function | false
 ---@param ... any
 ---@return ...
-local function triggerServerCallback(_, event, delay, cb, ...)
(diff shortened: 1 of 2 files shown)


thelindat:
The original plan was to check the size of the payload and use latent events automatically if deemed too large.

Unfortunately latent events have had a number of issues and numerous cfx devs have recommended against it's usage. Last I asked they were still discouraged but I need to confirm if they work or not; last I heard there was a memory issue, but that's still preferable to network hitches.

Lastly, if and when I implement this I intend to work on a new RPC class to supersede callbacks.

PR #657, opened 2024-10-22
feat(context): Added onMenu field option

AstroNext:
call onMenu function if available when menu id set to context menu option.


LukeWasTakenn (commented):
What exactly is the use case for something like this? To call a function when the menu is opened?

AstroNext:
When navigating from one menu to another, it's sometimes necessary to retain the arguments from the previous menu. For example, to edit the name or other details of a job rank, we first select the rank (assuming the menus are pre-defined and the transition occurs through the menu system not onSelect or event). At this point, it's essential to store this rank so that we can edit it.

PR #658, opened 2024-10-22
feat(radial): added canSelect functin and onEvent / onExport field

AstroNext:
added function to check player can select radial menu item or not.
added onEvent field to trigger event when player select radial item


Skryptific:
Thank you for the pull request, however it introduces a lot of duplicate code. the `canSelect `part could be done in a one-liner for example. It would be nice if you could take care of that.

AstroNext:
I have added `onExport` to use `export` functionality, similar to `onSelect` with `string` type.
I believe the `onSelect` with `string` type for `export` functionality should be removed as soon as possible.

PR #667, opened 2024-11-09
fix(timer): restart can prevent executing onEnd function

Frowmza:
If we want to restart a timer, previous state would disable the variable that triggers the OnEnd function by default.

### DIFF
--- a/imports/timer/shared.lua
+++ b/imports/timer/shared.lua
@@ -92,8 +92,13 @@ function timer:isPaused()
     return self.private.paused
 end
 
-function timer:restart(async, triggerOnEnd)
-    self:forceEnd(triggerOnEnd)
+function timer:restart(async, onEnd)
+    self:forceEnd(not not onEnd)
+
+    if type(onEnd) == 'function' then
+        self.private.onEnd = onEnd
+    end
+
     Wait(0)
     self.private.currentTimeLeft = self.private.initialTime
     self.private.startTime = 0
--- a/imports/timer/shared.lua
+++ b/imports/timer/shared.lua
@@ -93,10 +93,9 @@ function timer:isPaused()
 end
 
 function timer:restart(async)
-    self:forceEnd(false)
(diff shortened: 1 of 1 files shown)


jag3dagster:
Seems like a having a definitive one way or the other isn't ideal. An optional parameter that lets users toggle the trigger of the onEnd function. With the default being false to prevent a breaking change.

Frowmza:
agree, should be good!

imports/timer/shared.lua
```diff
@@ -92,8 +93,13 @@ function timer:isPaused()
...
 end
 
-function timer:restart(async)
-    self:forceEnd(false)
+function timer:restart(async, onEnd)
+    self:forceEnd(not not onEnd)
+
+    if type(onEnd) == 'function' then
+        self.private.onEnd = onEnd
+    end
```
jag3dagster, line 101: I don't think dynamically adding an onEnd function during a reset makes a lot of sense. What would this use case actually be?

@thelindat what are your thoughts on allowing an onEnd function to be added during a reset? Personally it feels like a really confusing user experience / flow.

imports/timer/shared.lua
```diff
@@ -92,8 +93,13 @@ function timer:isPaused()
     return self.private.paused
 end
 
-function timer:restart(async)
-    self:forceEnd(false)
+function timer:restart(async, onEnd)
+    self:forceEnd(not not onEnd)
```
jag3dagster, line 97: I'm assuming this double `not` flip is to handle the function case, but I think there is a better way to handle this if we do end up wanting the onEnd to accept a function.

thelindat:
I have a different idea to resolve this problem.

imports/timer/shared.lua
Frowmza: yeah just to pass it as `boolean` and not `function`. But let see what linden got!

PR #668, opened 2024-11-09
fix(callback): allow type table for cb

loaf-scripts:
Makes it so you can pass callbacks from RegisterNUICallback etc

### DIFF
--- a/imports/callback/client.lua
+++ b/imports/callback/client.lua
@@ -74,7 +74,7 @@ lib.callback = setmetatable({}, {
         else
             local cbType = type(cb)
 
-            assert(cbType == 'function', ("expected argument 3 to have type 'function' (received %s)"):format(cbType))
+            assert(cbType == 'function' or cbType == 'table', ("expected argument 3 to have type 'function' (received %s)"):format(cbType))
         end
 
         return triggerServerCallback(_, event, delay, cb, ...)


jag3dagster:
The use case that you described doesn't make a lot of sense. Could you give a more complete usage example for this functionality and how you imagine it working.

loaf-scripts:
An example would be to get a presigned URL to upload an image to Fivemanage.

Example client code:
```lua
RegisterNUICallback("getPresignedUrl", function(body, cb)
    lib.callback("getPresignedUrl", false, cb)
end)
```
With the latest version of ox_lib, this results in the following error:
`error during NUI callback getPresignedUrl: @ox_lib/imports/callback/client.lua:77: expected argument 3 to have type 'function' (received table)`

Another example would be an export that takes `cb` as an argument, e.g:
```lua
exports("GetData", function(cb)
    lib.callback("getData", false, cb)
end)
```
Currently, that would return the following error: 
image

jag3dagster:
Hm based on this commit 2909cb5 it looks like @thelindat is against this sort of functionality. I'll leave it up to him to decide if he thinks this use case is an exception.

thelindat:
The explained issue doesn't really make sense since nui callbacks aren't tables; so this is really just a fix for function references (still kind of weird/unnecessary usage tbh). What actually needs to be checked here is if it's a function or a table with a __call method.

PR #675, opened 2024-12-01
fix(resource/menu): prevent opening empty menu

BerkieBb:
Opening an empty menu always leads to an error and the nui code breaking. This should prevent most cases (as there can still be people that send empty tables in the options).


BerkieBb:
Yes, this is a thing that should be handled, I've unfortunately done this myself on accident a lot and spent too much time figuring out why it doesn't open, only for the nui code to be broken and a restart fixing it.

thelindat:
We probably need to look into resolving the underlying issue of the app breaking when there's an error, assuming that isn't a difficult task.

PR #679, opened 2024-12-12
tweak(resource/cache): update seat cache on vehicle change

mahanmoulaei:
With this change, seat cache event listeners will be triggered all the time when the player's vehicle changes. This is useful in scenarios where the player doesn't exit the vehicle, but we switch their vehicle and warp them into the same seat index, and we want to listen to the seat event to perform specific actions. Previously, seat listeners were not being triggered because the cached seat value stayed the same.


jag3dagster (commented):
Seems like a reasonable scenario to cover and the code looks straightforward enough @thelindat

Issue #548, opened 2024-04-16
[Bug] Zone creator return incorrrect value


Qotrek:
**Describe the bug**
The zone creator sometimes returns a value with commas instead of dots.

**To Reproduce**
1. Type command 'zone box'
2. Create zone
3. Save in array

**Expected behavior**
dots instead of commas

**Screenshots**
![image](https://github.com/overextended/ox_lib/assets/49820993/ffb04008-16f3-4d59-90a5-3ac10d8528c5)

LukeWasTakenn:
This might be due to regional formatting in Windows, where depending on the set region it will format numbers with `,` instead of `.`

PR #683, opened 2024-12-17
fix(resource/zonecreator): Fixed formatting vectors

Qotrek:


jag3dagster:
Is this to handle comma notation for decimal numbers?

jag3dagster:
Ah I see, it looks like this is a fix for #548

jag3dagster:
This seems like a reasonable enough workaround, but I have a feeling it would need to be handled on the createZone shape variants, not just box zones

PR #686, opened 2025-01-07
fix(marker): logical error when setting faceCamera

NotSomething0:
Due to the way boolean values were being assigned, faceCamera would always evaluate to true because of its default value. This change fixes the logical error and ensures boolean properties are properly set.

### DIFF
--- a/imports/marker/client.lua
+++ b/imports/marker/client.lua
@@ -5,9 +5,6 @@ local defaultRotation = vector3(0, 0, 0)
 local defaultDirection = vector3(0, 0, 0)
 local defaultColor = { r = 255, g = 255, b = 255, a = 100 }
 local defaultSize = { width = 2, height = 1 }
-local defaultBobUpAndDown = false
-local defaultFaceCamera = true
-local defaultRotate = false
 local defaultTextureDict = nil
 local defaultTextureName = nil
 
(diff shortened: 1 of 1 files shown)


jag3dagster:
This looks like a similar logic error as #672. Could you update your logic to match the formatting of that PR with a single line instead of these additional if blocks please?

NotSomething0:
Updated! I left the boolean check instead of just checking for nil since DrawMarker expects a boolean and the game doesn't care what you pass as a truthy value as long as it isn't false.

imports/marker/client.lua
```diff
@@ -152,9 +149,9 @@ function lib.marker.new(options)
   self.height = options.height or defaultSize.height
   self.rotation = options.rotation or defaultRotation
   self.direction = options.direction or defaultDirection
-  self.bobUpAndDown = options.bobUpAndDown or defaultBobUpAndDown
-  self.faceCamera = options.faceCamera or defaultFaceCamera
-  self.rotate = options.rotate or defaultRotate
+  self.bobUpAndDown = type(options.bobUpAndDown) ~= 'boolean' or options.bobUpAndDown
+  self.faceCamera = type(options.faceCamera) ~= 'boolean' or options.faceCamera
+  self.rotate = type(options.rotate) ~= 'boolean' or options.rotate
```
jag3dagster, line 154:
```lua
-- Default should be false
self.bobUpAndDown = type(options.bobUpAndDown) == 'boolean' and options.bobUpandDown
-- Default should be true
self.faceCamera = type(options.faceCamera) ~= 'boolean' or options.faceCamera
-- Default should be false
self.rotate = type(options.rotate) == 'boolean' and options.rotate
```

NotSomething0: Fixed I wasn't sure if the original default values were still wanted.

jag3dagster (commented):
Looks good to me @thelindat

Issue #682, opened 2024-12-16
[Bug] Removing and adding zones at runtime causes possible error [lib.zone]


sModsk:
When removing lib.zone uses zone.id as key for all tables where it removes zone. But tables enteringZones and exitingZones use different indexing. This causes possible conflict when wrong zone is removed from those tables.  It also breaks the iteration if the index is removed in between.

```lua
local function removeZone(self)
    Zones[self.id] = nil
    insideZones[self.id] = nil
    enteringZones[self.id] = nil
    exitingZones[self.id] = nil
end

...

if not zone.insideZone then
    zone.insideZone = true
    if zone.onEnter then
        enteringSize += 1
        enteringZones[enteringSize] = zone
    end

    if zone.inside or zone.debug then
        insideZones[zone.id] = zone
    end
end

...

for i = 1, enteringSize do
  enteringZones[i]:onEnter()
end
```

**Possible solution** would be using zone.id as key for all tables. And re-create table as sorted array by distance if triggering exit/enter methods in distance order is important. From an optimization perspective it has a little effect really.  

```lua
local function removeZone(self)
    Zones[self.id] = nil
    insideZones[self.id] = nil
    if enteringZones[self.id] then enteringSize -=1 end
    if exitingZones[self.id] then exitingSize -=1 end
    enteringZones[self.id] = nil
    exitingZones[self.id] = nil
end

local function getSortedZones(zoneTable, sortAscending)
    local sortedZones = {}

    for _, zone in pairs(zoneTable) do
        table.insert(sortedZones, zone)
    end

    table.sort(sortedZones, function(a, b)
        if sortAscending then
            return a.distance < b.distance
        else
            return a.distance > b.distance
        end
    end)

    return sortedZones
end


CreateThread(function()
    while true do
        local coords = GetEntityCoords(cache.ped)
        cache.coords = coords

        for _, zone in pairs(Zones) do
            zone.distance = #(zone.coords - coords)
            local radius, contains = zone.radius, nil

            if radius then
                contains = zone.distance < radius
            else
                contains = glm_polygon_contains(zone.polygon, coords, zone.thickness / 4)
            end

            if contains then
                if not zone.insideZone then
                    zone.insideZone = true

                    if zone.onEnter then
                        enteringSize += 1
                        enteringZones[zone.id] = zone
                    end

                    if zone.inside or zone.debug then
                        insideZones[zone.id] = zone
                    end
                end
            else
                if zone.insideZone then
[...]

PR #688, opened 2025-01-17
fix(client/zones): Prevent removal of zones mid-loop

jag3dagster:
Proposed fix for #682 without completely rewriting the structure / indexing of the entering / exiting zones tables.

Lua is inherently not thread-safe, but this adds an attempt at catching the zone loop between iterations when trying to remove an item.

Also removing the `enteringZones` and `exitingZones` update calls as this change should make it irrelevant and they weren't using proper indices anyways.


jag3dagster:
This is made irrelevant by 356e560

PR #698, opened 2025-01-26
Scaleform class

CodexisPhantom:
This add a scaleform class for a better management and usage of scaleforms.

At first it was a PR from Mycroft-Studios intended for qbx_core, but after discussions with the team we felt that this code would fit better in ox_lib so chatty asked for someone to create a PR.

### DIFF
--- a/imports/scaleform/client.lua
+++ b/imports/scaleform/client.lua
@@ -1,0 +1,229 @@
+---@class renderTargetTable
+---@field name string
+---@field model string|number
+
+---@class detailsTable
+---@field name string
+---@field fullScreen? boolean
+---@field x? number
+---@field y? number
+---@field width? number
+---@field height? number
+---@field renderTarget? renderTargetTable
+
+---@class Scaleform : OxClass
+---@field scaleform number
+---@field draw boolean
+---@field target number
+---@field targetName string
+---@field handle number
+---@field fullScreen boolean
+lib.scaleform = lib.class('Scaleform')
+
+--- Converts the arguments into data types usable by scaleform
+---@param argsTable table
+local function convertArgs(argsTable)
+    for i=1, #argsTable do
+        local arg = argsTable[i]
+        if type(arg) == 'string' then
+            ScaleformMovieMethodAddParamPlayerNameString(arg)
+        elseif type(arg) == 'number' then
+            if math.type(arg) == 'integer' then
+                ScaleformMovieMethodAddParamInt(arg)
+            else
+                ScaleformMovieMethodAddParamFloat(arg)
+            end
+        elseif type(arg) == 'boolean' then
+            ScaleformMovieMethodAddParamBool(arg)
+        else
+            error(('Unsupported Parameter type [%s]'):format(type(arg)))
+        end
+    end
+end
+
+---@param type 'boolean' | 'integer' | 'string'
+---@return boolean | integer | string
+---@description Awaits the return value, and converts it to a usable data type
+local function retrieveReturnValue(type)
+    local result = EndScaleformMovieMethodReturnValue()
+
+    lib.waitFor(function()
+        if IsScaleformMovieMethodReturnValueReady(result) then
+            return true
+        end
+    end, "Failed to retrieve return value", 1000)
+
+    if type == "integer" then
+        return GetScaleformMovieMethodReturnValueInt(result)
+    elseif type == "boolean" then
+        return GetScaleformMovieMethodReturnValueBool(result)
+    else
+        return GetScaleformMovieMethodReturnValueString(result)
+    end
+end
+
+---@param details detailsTable | string
+---@return nil
+---@description Create a new scaleform class
+function lib.scaleform:constructor(details)
+    details = type(details) == "table" and details or {name = details}
+
+    local scaleform = lib.requestScaleformMovie(details.name)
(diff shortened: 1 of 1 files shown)


imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+function lib.scaleform:constructor(details)
+    details = type(details) == "table" and details or { name = details }
+
+    local scaleform = lib.requestScaleformMovie(details.name)
+    if not scaleform then
+        return error(('Failed to request scaleform movie - [%s]'):format(details.name))
+    end
+
+    self.handle = scaleform
+    self.draw = false
```
thelindat, line 79: Please use more descriptive property names like `isDrawing` or such.
You could also use private properties or otherwise mark the property as private for LLS.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+    else
+        return GetScaleformMovieMethodReturnValueString(result)
+    end
+end
+
+---@param details detailsTable | string
+---@return nil
+---@description Create a new scaleform class
+function lib.scaleform:constructor(details)
+    details = type(details) == "table" and details or { name = details }
```
thelindat, line 71: You've mixed `'` and `"` for strings in a number of places. We generally use `'`.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+---@return any
+---@description Call a scaleform function, with optional args or return value.
+function lib.scaleform:callMethod(name, args, returnValue)
+    if not self.handle then
+        return error('Scaleform handle is nil')
+    end
+
+    if type(args) ~= 'table' then
+        return error('Args must be a table')
+    end
```
thelindat, line 104: These error messages could be formatted a little better. I've used a typeError function in some other parts of ox_lib.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+
+    BeginScaleformMovieMethod(self.handle, name)
+
+    if args then
+        convertArgs(args)
```
thelindat, line 109: `args` cannot be falsey, so this if statement is redundant.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+            ScaleformMovieMethodAddParamBool(arg)
+        else
+            error(('Unsupported Parameter type [%s]'):format(argType))
+        end
+    end
+end
+
+---@param expectedType 'boolean' | 'integer' | 'string'
+---@return boolean | integer | string
+---@description Awaits the return value, and converts it to a usable data type
```
thelindat, line 48: No need for `@description` here and in other places.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+
+    local scaleform = lib.requestScaleformMovie(details.name)
+    if not scaleform then
+        return error(('Failed to request scaleform movie - [%s]'):format(details.name))
+    end
+
+    self.handle = scaleform
+    self.draw = false
+
+    self.fullScreen = details.fullScreen ~= nil and details.fullScreen or true
```
thelindat, line 81: This will always be `true`.

imports/scaleform/client.lua
```diff
@@ -0,0 +1,231 @@
...
+    self.width = details.width or 0
+    self.height = details.height or 0
+
+    if details.renderTarget then
+        self:setRenderTarget(details.renderTarget.name, details.renderTarget.model)
+    end
+end
+
+---@param name string
+---@param args? table
```
thelindat, line 93: Should not be marked as optional, as your type check ensures it is a table. You should also clarify that it's an array-like table and which types are valid - `(number | string | boolean)[]`

thelindat (changes requested):

imports/scaleform/client.lua
CodexisPhantom: I corrected the code because args can be false. Here is a modified code from qbx_police where I tested the scaleform class.
```lua
local function setupIntructionalScaleform()
    if not currentScaleform then
        return
    end

    currentScaleform:callMethod('CLEAR_ALL')
    currentScaleform:callMethod('SET_CLEAR_SPACE', { 200 })
    currentScaleform:callMethod('SET_DATA_SLOT', { 1, GetControlInstructionalButton(1, 177, true), 'Fermer la caméra' })
    currentScaleform:callMethod('DRAW_INSTRUCTIONAL_BUTTONS')
    currentScaleform:callMethod('SET_BACKGROUND_COLOUR', { 0, 0, 0, 80 })
end
```

imports/scaleform/client.lua
```diff
@@ -0,0 +1,215 @@
...
+
+---@param details detailsTable | string
+---@return nil
+function lib.scaleform:constructor(details)
+    details = type(details) == 'table' and details or { name = details }
+
+    local scaleform = lib.requestScaleformMovie(details.name)
+
+    self.sfHandle = scaleform
+    self.private.isDrawing = false
```
thelindat, line 74: You've set isDrawing as private, but other references to the variable don't check the private table.

CodexisPhantom: Sorry I pushed without checking 😅

PR #703, opened 2025-02-01
refactor(relativeCoords): allow passing full rotation

dolutattoo:
Allow passing a full rotation (vector3) to get the relative coords.
Usefull if the entity is rotated on other axis than just vertical.
This prevent to use a vector4 as first arg though.


thelindat:
> This prevent to use a vector4 as first arg though.

Check if coords.w exists, rotation is a vector3, and offset is defined.

dolutattoo:
> > This prevent to use a vector4 as first arg though.
> 
> Check if coords.w exists, rotation is a vector3, and offset is defined.

Not sure if it's good to rename params in overlads but at least it's explicit like this

Issue #593, opened 2024-06-10
[Suggestion] ox print in cron


artur-michalak:
**Is your feature request related to a problem? Please describe.**
Ox_lib has a cool console printing mechanism. Maybe it would be useful to use lib.print.debug in Cron as a default debug mechanism?

**Describe the solution you'd like**
Use in Cron lib.print.debug when no debug parameter is specified.

**Additional context**
It might also be worth adding an export on the `resourcePrintLevel`.
https://github.com/overextended/ox_lib/blob/01334799f00499544c34abd0ba9d560ff855ef0d/imports/print/shared.lua#L18

PR #705, opened 2025-02-05
feat(server/cron): add allowExpiredTasks to execute overdue tasks

st860923:
Introduces a new `allowExpiredTasks` option for cron jobs, allowing scheduled tasks
to execute even if their scheduled time has already passed.

Previously, tasks that were slightly overdue (due to server hitches or thread
delays) were skipped, leading to inconsistencies in scheduling. With this change,
if `allowExpiredTasks = true`, tasks can still execute when overdue.

Changes:
- Added `allowExpiredTasks` as an optional parameter in `lib.cron.new`.

### DIFF
--- a/imports/cron/server.lua
+++ b/imports/cron/server.lua
@@ -27,6 +27,7 @@ setmetatable(currentDate, {
 ---@field id number
 ---@field debug? boolean
 ---@field lastRun? number
+---@field allowExpiredTasks? boolean
 
 ---@class OxTask : OxTaskProperties
 ---@field expression string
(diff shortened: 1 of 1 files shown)


thelindat:
Not really a suitable fix.

@juddisjudd Do you think we can remove this check outright now? It was there to cover for some edge cases where jobs were incorrectly scheduled.

juddisjudd:
> Not really a suitable fix.
> 
> @juddisjudd Do you think we can remove this check outright now? It was there to cover for some edge cases where jobs were incorrectly scheduled.

I will take a look and see what would be best for this.

juddisjudd:
@thelindat I think maybe this would be a better approach:
Add a maxDelay option:
`---@field maxDelay? number Maximum allowed delay in seconds before skipping (default: 0)`

```lua
if sleep < 0 then
    if not self.maxDelay or -sleep > self.maxDelay then
        return self:stop(self.debug and ('scheduled time expired %s seconds ago'):format(-sleep))
    end
    
    if self.debug then
        print(('Task %s is delayed by %s seconds, executing now'):format(self.id, -sleep))
    end
    
    sleep = 0
end
```

Example usage then setting something like this:
`maxDelay = 30` for a 30 second grace period or something

thelindat:
Sure we can go with that, maybe default to 1 second.

While we're at it, it would be good to swap to lib.print.debug and default debug to true if printLevel is set to debug. See https://github.com/overextended/ox_lib/issues/593

juddisjudd:
@st860923 would you like to update your PR with my suggestion or would you prefer that I do this?

st860923:
You can proceed with your suggested approach. Thank you!

PR #707, opened 2025-02-05
web build index error fix

L1quidSn4ke:
File moved so I changed the code to reflect the change

### DIFF
--- a/fxmanifest.lua
+++ b/fxmanifest.lua
@@ -16,15 +16,14 @@ dependencies {
     '/onesync',
 }
 
-ui_page 'web/build/index.html'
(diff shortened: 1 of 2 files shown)


HeyyCzer:
This file was not moved. You probably did not build the resource.

If you has no experience I recommend to download the built version from the [Releases](https://github.com/overextended/ox_lib/releases) page. If it's not the case, follow the guide from the [Documentation - Building the UI](https://overextended.dev/ox_lib#building-the-ui)

Ehbw:
This file hasn't been moved. You need to build the UI. refer to the links above

PR #716, opened 2025-02-27
fix(imports/zones): average z coordinate of non planar polys

JanBlk:
I noticed that when creating a polyzone with points that have very similar, but not equal, z coordinates (so if they all round to the same number), the previous averaging logic threw an error, it was also overcomplicated.

### DIFF
--- a/imports/zones/client.lua
+++ b/imports/zones/client.lua
@@ -265,6 +265,18 @@ local function convertToVector(coords)
     return coords
 end
 
+local function getAverageZ(vecs)
+    if not vecs or type(vecs) ~= 'table' or #vecs == 0 then return 0 end
+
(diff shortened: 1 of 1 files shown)


tom-osborne:
This function does indeed calculate the average value of all the z-coordinates.

However, this does produce a different result that the current ox implementation.

I created a simple test [here](http://tpcg.io/_5ZOZO4). 

When you provide a list of points without a duplicate z-coord, the current ox implementation breaks. But when a duplicate z-coord is present, it doesn't break, but also doesn't calculate the average correctly.

@thelindat @DokaDoka I think the implementation in this PR (correctly calculating the average z-coord) produces the desired result. Would you agree?

PR #728, opened 2025-03-10
feat(print): add convar change listener

CENSOR1337:
add convar change listener to print module, This would allow to set their convar at their runtime and see information without restarting resource.

### DIFF
--- a/imports/print/shared.lua
+++ b/imports/print/shared.lua
@@ -14,12 +14,10 @@ local levelPrefixes = {
     '^4[VERBOSE]',
     '^6[DEBUG]',
 }
-local convarNames = {
-    resource = 'ox:printlevel:' .. cache.resource,
-    global = 'ox:printlevel',
-}
+local convarGlobal = 'ox:printlevel'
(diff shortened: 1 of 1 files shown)


imports/print/shared.lua
```diff
@@ -14,15 +14,27 @@ local levelPrefixes = {
     '^4[VERBOSE]',
     '^6[DEBUG]',
 }
-
-local resourcePrintLevel = printLevel[GetConvar('ox:printlevel:' .. cache.resource, GetConvar('ox:printlevel', 'info'))]
+local convarNames = {
+    resource = 'ox:printlevel:' .. cache.resource,
+    global = 'ox:printlevel',
+}
```
thelindat, line 20: Just have two variables rather than a table.

thelindat:
Should probably check if that native exists too, otherwise the server dependency needs to be updated in fxmanifest.

CENSOR1337:
> Should probably check if that native exists too, otherwise the server dependency needs to be updated in fxmanifest.

Oh that's right, 

Thanks for the suggestion

PR #734, opened 2025-03-16
feat(menu): Add isVisible field to the option table.

fnbar0:
Works similarly to ox_target's canInteract function, running isVisible function for each menu option (if defined) and determining its visibility based on its boolean return value.

### DIFF
--- a/resource/interface/client/menu.lua
+++ b/resource/interface/client/menu.lua
@@ -18,6 +18,7 @@ local openMenu
 ---@field defaultIndex? number
 ---@field args? {[any]: any}
 ---@field close? boolean
+---@field isVisible? fun(): boolean
 
 ---@class MenuProps
 ---@field id string
@@ -48,6 +49,12 @@ function lib.showMenu(id, startIndex)
     local menu = registeredMenus[id]
     if not menu then
         error(('No menu with id %s was found'):format(id))
+    end
+
+    for i = #menu.options, 1, -1 do
+        if menu.options[i].isVisible and not menu.options[i].isVisible() then
+            table.remove(menu.options, i)
+        end
     end
 
     if table.type(menu.options) == 'empty' then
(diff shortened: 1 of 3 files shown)


resource/interface/client/menu.lua
```diff
@@ -50,6 +51,12 @@ function lib.showMenu(id, startIndex)
         error(('No menu with id %s was found'):format(id))
     end
 
+    for i = #menu.options, 1, -1 do
+        if menu.options[i].isVisible and not menu.options[i].isVisible() then
+            table.remove(menu.options, i)
+        end
+    end
```
RAlmada, line 58: This will technically do what you think it will... until an option has been deemed not visible, then that option will never been seen again unless the menu is re-registered with that option. `menu` is a reference to the underlying stored table, not a clone, if something is removed it won't appear again the next time you go to open the menu. If you are just going to need to re-register the menu anyways, this is wasted computation time when you could just have logic when you are setting up the menu to just not include the option as is.

RAlmada:
If anything, this functionality would likely be better as a flag that the UI then handles if it is marked as not visible. But in its current form I don't believe this functionality is desired.

@LukeWasTakenn @thelindat I pass it to you guys for final decision / comments.

resource/interface/client/menu.lua
jag3dagster, line 58: This will technically do what you think it will... until an option has been deemed not visible, then that option will never been seen again unless the menu is re-registered with that option. menu is a reference to the underlying stored table, not a clone, if something is removed it won't appear again the next time you go to open the menu. If you are just going to need to re-register the menu anyways, this is wasted computation time when you could just have logic when you are setting up the menu to just not include the option as is.

jag3dagster:
If anything, this functionality would likely be better as a flag that the UI then handles if it is marked as not visible. But in its current form I don't believe this functionality is desired.

@LukeWasTakenn @thelindat I pass it to you guys for final decision / comments.

resource/interface/client/menu.lua
fnbar0: You're right, just replaced it with the flag you've mentioned.

PR #737, opened 2025-03-19
feat(web): validate if primary color and shade are available in mantine

RrybaN:
If you misspell or enter an invalid mantine theme color, the UI will not render. Added validation to prevent this.


LukeWasTakenn:
Are custom colors also loaded into the default theme?

RrybaN:
AFAIK, custom colors are not automatically added to `DEFAULT_THEME`. I think you need to define them in a custom theme and provide it to `MantineProvider` for them to be recognized. Correct me if I'm wrong.

PR #745, opened 2025-04-10
fix: check interval before calling a cb

ESK0:
This PR fixes an issue where once a Interval is cleared the CB is called twice.

Repro:
image
image


jag3dagster (approved):
LGTM based off the conversation had in book club

PR #751, opened 2025-04-24
refactor(lib): smarter module system with reload + deps

atoshit:
Replaces the default lib.require with an enhanced version supporting:
- Hot-reloading of modules (lib.reload, lib.reloadAll)
- Dependency injection with depends and init()
- Sandboxed environments for safer module execution
- Built-in support for module-level unit testing

Improves modularity, maintainability, and developer workflow across resources.

### DIFF
--- a/imports/require/shared.lua
+++ b/imports/require/shared.lua
@@ -7,39 +7,54 @@
 ]]
 
 local loaded = {}
+local moduleEnvs = {}
 local _require = require
+
+local defaultEnv = setmetatable({
+    print = print,
+    pairs = pairs,
+    ipairs = ipairs,
+    string = string,
+    table = table,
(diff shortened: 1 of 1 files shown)


imports/require/shared.lua
```diff
@@ -93,48 +98,32 @@ local function loadModule(modName, env)
     return nil, err or 'unknown error'
 end
 
----@alias PackageSearcher
```
Maximus7474, line 96: Why remove all the luadoc ?

Ehbw:
Breaks compatibility with any script using require (with the "sandboxing") the hot reloading implementation will not work in the way you believe it does and the rest of the code changes are not ideal especially removing comments and luadocs

PR #753, opened 2025-04-26
Fix paintType application for default and custom vehicle colors

MurilloGavron:
Previously, applying paintType1 and paintType2 (for Metallic, Matte, Chrome paint types) didn't work correctly.
Even when setting these values, the vehicle would ignore the desired paint type and fallback to default behavior (usually Metallic).

The issue was caused by improper handling of SetVehicleModColor_1/2, and a lack of distinction between custom RGB colors and default GTA color IDs. Also, it didn't properly clear custom colors when switching back to a stock paint ID.

This fix ensures that:

- If color1 or color2 are numbers (GTA default colors), they are applied correctly and paintType is set properly.
- If color1 or color2 are tables (custom RGB), they are used as custom colors and still apply paintType.

- ClearVehicleCustomPrimaryColour and ClearVehicleCustomSecondaryColour are correctly used to avoid color conflicts.

Now the paintType properly changes the paint finish (Metallic, Matte, Chrome) as intended.

### DIFF
--- a/resource/vehicleProperties/client.lua
+++ b/resource/vehicleProperties/client.lua
@@ -347,26 +347,42 @@ function lib.setVehicleProperties(vehicle, props, fixVehicle)
     end
 
     if props.color1 then
-        if type(props.color1) == 'number' then
+        if type(props.color1) == 'table' then
+            -- Cor personalizada (RGB)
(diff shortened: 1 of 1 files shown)


Ehbw:
Please remove or translate the comments to english

PR #756, opened 2026-04-28
feat(test): add lib.test unit-testing framework

Kenshiin13:
## What this adds

A unit testing framework, `lib.test`. Server-side, runs in the FiveM runtime, modeled on Vitest.

```lua
lib.test.describe('Inventory', function()
    lib.test.it('starts empty', function()
        lib.test.expect(Inventory:new():count('bread')):toBe(0)
    end)
end)
```

Full reference, setup instructions, matcher list, async patterns, run options, troubleshooting and a small cookbook live in [imports/test/README.md](https://github.com/Kenshiin13/ox_lib/blob/tdd/imports/test/README.md).

## Why

ox_lib has no test infrastructure today. Every consumer rolls their own or skips testing entirely. `lib.test` gives resource authors a first-class way to write unit tests against their server logic with a Vitest-style API: describe / it / expect, hooks, mocks, spies, async support with per-test timeouts, parameterized tests, and JSON output for CI.

## Setup for consumers

Declare a test directory in your fxmanifest and depend on ox_lib:

```lua
dependency 'ox_lib'
ox_test_dir 'tests'

server_scripts {
    '@ox_lib/init.lua',
    'server/your_code.lua',
}
```

Drop test files anywhere under `tests/`, named `*.test.lua`. From the server console:

```
oxtest                       # run ox_lib's bundled examples
oxtest my_resource           # run tests in my_resource
oxtest my_resource auth      # filter by substring
```

Or programmatically from another resource via the registered export:

```lua
local result = exports['my_resource']:runOxTests('auth', 'console')
```

## Feature surface

Everything below has at least one passing self-test.

- `describe`, `it`, `it.skip`, `it.only`, `it.each` (passing.lua, parameterized.lua, run_options.lua)
- `beforeEach`, `afterEach`, `beforeAll`, `afterAll` with outer-then-inner ordering, plus failure handling for each (hooks.lua)
- 16 matchers with chainable `.never` modifier (passing.lua, matchers.lua, assertion_errors.lua)
- Asymmetric matchers: `expect.any`, `expect.anything`, `expect.callable`, `expect.objectContaining`, `expect.arrayContaining` (matchers.lua)
- `lib.test.fn()` callable mocks with `mockReturnValue`, `mockImplementation`, `mockClear`, `mockReset` (mocks.lua)
- `lib.test.spy(obj, key)` with auto-restoration after each test, including warning path for inherited methods (spies.lua)
- Three async completion modes (return promise, `done()` callback, inline `Citizen.Await`) with per-test timeout race (async.lua, timeouts.lua)
- `lib.test.isCallable(v)` and `:toBeCallable()` for production code that needs to handle FiveM cross-resource function refs (callable tables, not `type() == 'function'`) (matchers.lua)
- Run options: filter by substring, bail on first failure, custom reporter, JSON reporter, default timeout override (run_options.lua)
- Recursive `*.test.lua` discovery from `ox_test_dir` manifest entries
- `runOxTests` export for cross-VM invocation
- Console reporter with Jest-style summary; JSON reporter for CI
- `runIsolated(setupFn, opts)` for self-tests that need to verify run-level behavior without disturbing the active run

## Self-test status

`oxtest` with no args runs every example file. All passing. Failure-message verification is done by wrapping deliberately-failing assertions in `:toThrow(pattern)`, so the suite stays green and any red line is a real regression.

The bundled `_demo_failures.lua` is intentionally not auto-loaded; load it manually with `lib.load` if you want to see what failure / timeout output looks like in the reporter.

## Changes to existing code

These are scoped, minimal, and necessary for the framework to work cleanly. None should affect existing consumers.

**`init.lua`** picks up a new manifest signal. After the existing `ox_lib '<name>'` auto-load loop it now also auto-loads `lib.test` in any consumer that declares at least one `ox_test_dir`. This is required because tests run in the consumer's Lua VM (where its globals exist), and the lazy autoloader would otherwise never fire if the consumer never directly references `lib.test`. Resources that don't declare `ox_test_dir` skip the load entirely.

**`fxmanifest.lua`** adds `imports/test/server.lua` to `server_scripts` so the `/oxtest` command registers at ox_lib startup. Same pattern as `imports/callback/server.lua` and `imports/getFilesInDirectory/server.lua` already in there.

**`imports/getFilesInDirectory/server.lua`** gets a recursion option and a few reliability fixes that came up while building discovery. Backward compatible: the existing two-argument call signature behaves exactly the same.

- Optional third boolean `recursive` (default false). When true, walks subdirectories.
- OS detection now inspects `GetResourcePath()` for drive letter or backslash instead of `os.getenv('OS')`. The env var isn't reliable in some FXServer environments where the parent process doesn't pass it through.
- Strips trailing `\r` from each line. Windows `io.popen` returns CRLF-terminated output, which broke end-of-string Lua patterns.
- Filters out `.` and `..` entries.
- Handles shells that return bare filenames (busybox / alpine) as well as full paths (Windows `dir /s`, unix `find`).

## How `lib.test` is laid out

Eight focused files in `imports/test/`, each under 450 lines.

- `server.lua` is the entry point. Requires the submodules, assembles `lib.test`, registers the `runOxTests` export in every VM that loads it, and registers the `/oxtest` command only inside ox_lib's own VM.
- `helpers.lua` is the only stateless module: `formatValue`, `deepEqual` (with cycle detection), `nodePath`, `newHooks`, `isCallable`, `indent`, `stripColors`.
- `registry.lua` holds the suite tree, current-context stack, and the only-mode flag. State is exposed on the module table so `runner` can swap it during `runIsolated`.
- `dsl.lua` is `describe` / `it` / hooks. Pure registration, no execution.
- `expect.lua` is every matcher, the chainable `.never` modifier, and the asymmetric matchers.
- `mock.lua` is `fn`, `spy`, `pendingRestores`, `restoreAll`. Spy restoration entries are pushed onto a module-level array and drained by the runner after every test.
- `runner.lua` walks the registry, races each test against its timeout, dispatches reporter callbacks, and provides `runIsolated`. The async detection uses `debug.getinfo(...).nparams` to distinguish sync / promise / done() bodies.
- `reporter.lua` is the console reporter (Jest-style summary), the JSON reporter, and an empty silent reporter.
- `discover.lua` reads `ox_test_dir` metadata, scans recursively via `lib.getFilesInDirectory`, loads each `*.test.lua` file into the calling VM. Also `register(path)` for explicit single-file loads.

## Out of scope

Considered but deliberately deferred. Documenting here so reviewers don't ask.

- Client-side testing. Doable, but skipped for v1; meaningful unit-testable logic in FiveM resources tends to live server-side.


thelindat:
I have to insist this be kept separate from ox_lib.

PR #757, opened 2026-04-28
fix(getNearbyPlayers): More reliable coord getter for players in vehicle

tugamars:
When a player is inside a vehicle GetEntityCoords returns the center of the vehicle, which is not necessarily the players coords in the world.

This PR improves the reliability of getNearbyPlayers by using the actual position of the player in the world.

### DIFF
--- a/imports/getClosestPlayer/client.lua
+++ b/imports/getClosestPlayer/client.lua
@@ -25,6 +25,10 @@ function lib.getClosestPlayer(coords, maxDistance, includePlayer)
 			local playerCoords = GetEntityCoords(playerPed)
 			local distance = #(coords - playerCoords)
 
+            if(GetVehiclePedIsIn(playerPed,false) ~= 0 ) then
+				playerCoords=GetWorldPositionOfEntityBone(playerPed,0);
+			end 
+
 			if distance < maxDistance then
 				maxDistance = distance
 				closestId = playerId
--- a/imports/getClosestPlayer/client.lua
+++ b/imports/getClosestPlayer/client.lua
@@ -14,7 +14,7 @@
 ---@return vector3? playerCoords
 function lib.getClosestPlayer(coords, maxDistance, includePlayer)
 	local players = GetActivePlayers()
-	local closestId, closestPed, closestCoords
+	local closestId, closestPed, closestCoords, closestVehicle
 	maxDistance = maxDistance or 2.0
 
 	for i = 1, #players do
@@ -24,9 +24,8 @@ function lib.getClosestPlayer(coords, maxDistance, includePlayer)
 			local playerPed = GetPlayerPed(playerId)
 			local playerCoords = GetEntityCoords(playerPed)
 
-            if GetVehiclePedIsIn(playerPed, false) ~= 0 then
-				playerCoords=GetWorldPositionOfEntityBone(playerPed, 0)
-			end 
+            local vehicle = GetVehiclePedIsIn(playerPed, false)
+            local playerCoords = vehicle == 0 and GetEntityCoords(playerPed) or GetWorldPositionOfEntityBone(playerPed, 0)
             
 			local distance = #(coords - playerCoords)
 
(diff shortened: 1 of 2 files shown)


unitysync (commented):
I sense this could possibly cause issues with large vehicles like trucks, since you could *technically* be far away from the chassis bone at the centre of the vehicle. Are you able to test your changes with a scenario like that with a few different large vehicles (`bus`, `phantom`, `dump`)?

tugamars:
> I sense this could possibly cause issues with large vehicles like trucks, since you could _technically_ be far away from the chassis bone at the centre of the vehicle. Are you able to test your changes with a scenario like that with a few different large vehicles (`bus`, `phantom`, `dump`)?

That is precisely what this aims to solve.

Currently (before update) the logic goes like this:
- If player is out of a vehicle it will obtain your coords correctly (GetEntityCoords)
- If player is in a vehicle, it will return the vehicles coords for that player - which may differ from the actual player location in the world

This is not really "too bad" with small vehicles because the offset from the actual player location to the vehicle center is small. 
But for large vehicles (like the ones you mentioned) this will cause issues.

New logic with the update:
- Inaltered for players outside of vehicle
- For players in a vehicle it will obtain the Peds absolute bone location in the World

So on small vehicles it will be a more accurate position (and therefor distance from the client that calculated it) and in big vehicles where it previously would return no one was nearby (due to being far from center of vehicle) it will return an accurate reading.

**This was tested with the bus and the limo vehicles**

imports/getClosestPlayer/client.lua
```diff
@@ -25,6 +25,10 @@ function lib.getClosestPlayer(coords, maxDistance, includePlayer)
 			local playerCoords = GetEntityCoords(playerPed)
 			local distance = #(coords - playerCoords)
 
+            if GetVehiclePedIsIn(playerPed, false) ~= 0 then
+				playerCoords=GetWorldPositionOfEntityBone(playerPed, 0)
+			end
```
thelindat: What's the origin of GetEntityCoords (on foot) compared to GetWorldPositionOfEntityBone? Can potentially just the latter if it doesn't throw off coordinates.

Otherwise, get the coords just once. Distance is also wrong here when the player is in a car.

tugamars: The difference is not much when you are stopped in place:
On Foot:
EntityCoords: vector3(-1019.160, -2890.663, 13.964)
Bone[0] Coords: vector3(-1019.144, -2890.665, 13.937)
Distance between EntityCoords and Bone[0]: 0.032

Sedan
EntityCoords: vector3(-1028.759, -2887.345, 13.229)
Bone[0] Coords: vector3(-1028.725, -2886.955, 13.353)
Distance between EntityCoords and Bone[0]: 0.411

Bus
EntityCoords: vector3(-1042.814, -2963.536, 13.946)
Bone[0] Coords: vector3(-1037.405, -2964.319, 14.195)
Distance between EntityCoords and Bone[0]: 5.471

-

When you are walking/running however it's not consistent:
Distance between EntityCoords and Bone[0]: 0.120
Distance between EntityCoords and Bone[0]: 0.116
Distance between EntityCoords and Bone[0]: 0.063
Distance between EntityCoords and Bone[0]: 0.113
Distance between EntityCoords and Bone[0]: 0.103
Distance between EntityCoords and Bone[0]: 0.113
Distance between EntityCoords and Bone[0]: 0.106

thelindat: In that case the following would be better. There would also be some value in returning the closest player's vehicleid since we're getting it anyway.
```lua
local vehicle = GetVehiclePedIsIn(playerPed, false)
local playerCoords = vehicle == 0 and GetEntityCoords(playerPed) or GetWorldPositionOfEntityBone(playerPed, 0)
```

You've also labeled the PR as getNearbyPlayers but only made a change for getClosestPlayer here; updating both would be sensible.

Legacy-TacticalGamingInteractive:
Excellent ty

PR #762, opened 2026-05-02
add server-side and client-side entity spawn wrappers and shared helper

TGIANN:
Add `lib.createObject`, `lib.createPed`, and `lib.createVehicle` for the server, plus a shared `lib.spawnEntity` helper used by both client and server wrappers. Each wrapper splits its body at module load time based on `cache.game` so that the correct CREATE_OBJECT/PED/VEHICLE native signature is used on FiveM vs RedM.

Server wrappers default `orphanMode` to 2 (KeepEntity). Server `createVehicle` uses `CreateVehicleServerSetter` (with `vehicleType`) on FiveM and the CFX `CreateVehicle` native on RedM, where `vehicleType`, `properties`, and `seat` are not exposed.

Existing client wrappers were refactored to delegate model loading to `lib.spawnEntity`, with a few client-side fixes folded in:
- pass heading as the float arg in `SetEntityHeading` for createObject
- replace undefined `netMissionEntity` with `isNetwork` in createPed
- drop duplicate `isNetwork` assert in createPed
- mark `isNetwork`, `netMissionEntity`, and `doorFlag` as optional in the type annotations to match runtime behavior


TGIANN:
docs: https://github.com/overextended/overextended.github.io/pull/273

fxmanifest.lua
```diff
@@ -43,6 +43,10 @@ client_scripts {
 server_scripts {
     'imports/callback/server.lua',
     'imports/getFilesInDirectory/server.lua',
+    'imports/spawnEntity/server.lua',
+    'imports/createObject/server.lua',
+    'imports/createPed/server.lua',
+    'imports/createVehicle/server.lua',
```
thelindat: These are unnecessary.

imports/createVehicle/client.lua
```diff
@@ -0,0 +1,82 @@
...
+---@alias SeatPosition
+---| -2 # SF_ANY
+---| -1 # SF_FrontDriverSide
+---| 0 # SF_FrontPassengerSide
+---| 1 # SF_BackDriverSide
+---| 2 # SF_BackPassengerSide
+---| 3 # SF_AltFrontDriverSide
+---| 4 # SF_AltFrontPassengerSide
+---| 5 # SF_AltBackDriverSide
+---| 6 # SF_AltBackPassengerSide
```
thelindat: It would be best to have a types file or have game types like this defined as part of fivem-lls-addon.

thelindat:
My biggest issues with this are
- code duplication
- type disparity (i.e. function declarations differ greatly between fivem/redm/fxserver)
- uses RPCs in some cases (almost never good)

I'd personally want to see classes used for any entity wrappers, e.g.

lib.entity as a base class inherited by lib.object, lib.ped, lib.vehicle with some common methods across server and client.

Networked entities from clients are also somewhat discouraged and potentially disabled on many servers, so focusing on server-created entities is more important.

TGIANN:
> My biggest issues with this are
> 
> * code duplication
> * type disparity (i.e. function declarations differ greatly between fivem/redm/fxserver)
> * uses RPCs in some cases (almost never good)
> 
> I'd personally want to see classes used for any entity wrappers, e.g.
> 
> lib.entity as a base class inherited by lib.object, lib.ped, lib.vehicle with some common methods across server and client.
> 
> Networked entities from clients are also somewhat discouraged and potentially disabled on many servers, so focusing on server-created entities is more important.

I've re edit everything. If everything is as you want and there are no issues, I can prepare the docs

test script
[tgiann-test.zip](https://github.com/user-attachments/files/27303350/tgiann-test.zip)

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+-- Capture the global `Entity()` state-bag accessor before our local class shadows it.
+local getEntityStateBag = Entity
+
+---Base class wrapping a CFX entity handle. Used directly via `lib.entity:new(handle)`
+---to wrap any pre-existing entity, or as the parent of `lib.object`, `lib.ped`, and
+---`lib.vehicle` for typed spawn wrappers.
+---@class Entity : OxClass
+---@field handle number Native entity handle.
+---@field script string Resource that created or wrapped this entity.
+local Entity = lib.class('Entity')
```
thelindat, line 18: You'll need to use `lib.entity` from the start. Assigning the value later results in private and protected fields giving errors. Same applies to the other classes of course.

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+]]
+
+-- Capture the global `Entity()` state-bag accessor before our local class shadows it.
+local getEntityStateBag = Entity
+
+---Base class wrapping a CFX entity handle. Used directly via `lib.entity:new(handle)`
+---to wrap any pre-existing entity, or as the parent of `lib.object`, `lib.ped`, and
+---`lib.vehicle` for typed spawn wrappers.
+---@class Entity : OxClass
+---@field handle number Native entity handle.
```
thelindat, line 16: Unfortunately there's no readonly annotation, but always good to tell people not to mess with values.

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+
+local IS_SERVER = IsDuplicityVersion()
+
+---@param handle number
+function Entity:constructor(handle)
+    local handleType = type(handle)
+    assert(handleType == 'number' and handle ~= 0, ('expected non-zero entity handle, got %s (%s)'):format(handleType, tostring(handle)))
+
+    self.handle = handle
+    self.script = GetInvokingResource() or cache.resource
```
thelindat, line 28: What's this for exactly?

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+
+---@return number
+function Entity:getModel()
+    return GetEntityModel(self.handle)
+end
+
+---Returns the entity's state bag.
+function Entity:getState()
+    return getEntityStateBag(self.handle).state
+end
```
thelindat, line 83: I'd scrap this for now, in case we do our own statebag handling (https://github.com/overextended/ox_lib/issues/778)

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+    if newHandle == 0 then return nil end
+
+    self.handle = newHandle
+
+    if heading then self:setHeading(heading) end
+    if data.rotation then self:setRotation(data.rotation) end
+
+    if IS_SERVER and cache.game ~= 'redm' then
+        self:setOrphanMode(data.orphanMode or 2)
+    end
```
thelindat, line 124: `cache.game` on the server will be 'fxserver'.

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+
+    return newHandle
+end
+
+---@protected
+---@param data table The cloned spawn data used for this respawn.
+function Entity:onAfterRespawn(data) end
+
+if IS_SERVER then
+    local allowClientServerEntityCreation = GetConvarInt('ox:allowClientServerEntityCreation', 0) == 1
```
thelindat, line 139: For now let's just focus solely on class wrappers and not implementing RPCs.

imports/object/server.lua
```diff
@@ -0,0 +1,65 @@
...
+---@param modelHash number
+---@param data ObjectInitServer
+---@return number handle
+function ObjectServer.spawn(modelHash, data)
+    if cache.game == 'redm' then
+        return CreateObject(modelHash, data.coords.x, data.coords.y, data.coords.z,
+            true, data.bScriptHostObj or false, data.dynamic or false, data.p7 or false, data.p8 or false)
+    end
+
+    return CreateObject(modelHash, data.coords.x, data.coords.y, data.coords.z, true, true, data.doorFlag or false)
```
thelindat, line 57: CreateObject is an RPC; the server needs to use [CreateObjectNoOffset](https://docs.fivem.net/natives/?_0x58040420).

imports/entity/shared.lua
```diff
@@ -0,0 +1,270 @@
...
+    Copyright © 2025 Linden <https://github.com/thelindat>
+]]
+
+-- Capture the global `Entity()` state-bag accessor before our local class shadows it.
+local getEntityStateBag = Entity
+
+---Base class wrapping a CFX entity handle. Used directly via `lib.entity:new(handle)`
+---to wrap any pre-existing entity, or as the parent of `lib.object`, `lib.ped`, and
+---`lib.vehicle` for typed spawn wrappers.
+---@class Entity : OxClass
```
thelindat, line 15: This should probably be `GameEntity`.

thelindat (commented):
This is usable but I might need to do some planning for a future implementation to ensure its fully fit for purpose. This can also work into a proposal for a custom statebag interface and would need a solid TS implementation too.

Things like resource management/cleanup, grid handling, etc. would probably be necessary for any (local) client entities too.

thelindat:
New classes and create helpers have been added. If you want to add additional methods and features, please open a new PR.

Issue #549, opened 2024-04-18
[Bug] lib.class private table returns not expected value


deFuzz77:
**Describe the bug**
I have problem with private table in lib.class, acctualy when I call method from the same method, the private table returns null.

**To Reproduce**
I wrote simple code to reproduce this.

```
local Inventory = lib.class('Inventory')

function Inventory:constructor()
    self.private.items = { { name = 'bread', count = 1 }, { name = 'water', count = 2 } }
end

function Inventory:getItem(rec)
    print(json.encode(self.private.items))
    rec = rec or 0
    if rec == 1 then return end
    rec += 1
    return self:getItem(rec)
end

local playerInventory = Inventory:new()
playerInventory:getItem()

--[[
    returns:
        --> [{"count":1,"name":"bread"},{"count":2,"name":"water"}]
        --> null
]]
```

**Expected behavior**
I expected the private table return its data.

PR #774, opened 2026-05-05
fix(class): close anonymous-function bypass of private field gate

Kenshiin13:
### Description

`OxClass` instances expose a `private` table whose metatable is supposed to allow reads and writes only from the class's own methods. The implementation gates access by checking `debug.getinfo(2, 'n').namewhat` of the immediate caller:

```lua
__index = function(self, index)
    local di = getinfo(2, 'n')
    if di.namewhat ~= 'method' and di.namewhat ~= '' then return end
    return private[index]
end,
__newindex = function(self, index, value)
    local di = getinfo(2, 'n')
    if di.namewhat ~= 'method' and di.namewhat ~= '' then
        error(("cannot set value of private field '%s'"):format(index), 2)
    end
    private[index] = value
end
```

The intended allowance was `'method'`. The `''` clause was added to handle internal cases where the caller's namewhat is empty, but it has the side effect of also allowing every anonymous function, every function called via `pcall`, every C-function intermediary, and the main chunk. Three keystrokes wrap any access in an anonymous function and bypass the gate entirely.

### Reproduction

```lua
local Test = lib.class('Test')

function Test:constructor()
    self.private.secret = 42
end

function Test:getSecret()
    return self.private.secret
end

local obj = Test:new()

-- Method access (intended): allowed
print(obj:getSecret())                                              -- 42

-- Global function: namewhat='global', denied
function steal()
    return obj.private.secret
end
print(steal())                                                      -- nil

-- Anonymous IIFE: namewhat='', allowed (bypass)
print((function() return obj.private.secret end)())                 -- 42

-- pcall: same path, allowed (bypass)
print((select(2, pcall(function() return obj.private.secret end)))) -- 42

-- Anonymous write also bypasses
;(function() obj.private.secret = 'pwned' end)()
print(obj:getSecret())                                              -- pwned
```

Any code that puts invariant-bearing state in `self.private` is wrong about its access control. The anonymous-function pattern is trivially constructable, and `pcall`-wrapping covers the slightly less ergonomic case where someone wants to silence errors from the access.

### Fix

Replace the immediate-caller `namewhat` check with a stack walk that allows access if any frame in the call chain is a method. This blocks the external bypasses (anonymous IIFE, pcall, global/local/upvalue access) while keeping legitimate intra-method patterns working. Most importantly `pcall(function() self.private.x = ... end)` inside a method body, which is a normal way to wrap setup code.

```lua
local function calledFromMethod()
    local level = 3
    while true do
        local di = getinfo(level, 'n')
        if not di then return false end
        if di.namewhat == 'method' then return true end
        level += 1
    end
end

setmetatable(obj.private, {
    __metatable = 'private',
    __tostring = void,
    __index = function(self, index)
        if not calledFromMethod() then return end
        return private[index]
    end,
    __newindex = function(self, index, value)
        if not calledFromMethod() then
            error(("cannot set value of private field '%s'"):format(index), 2)
        end
        private[index] = value
    end
})
```

The walk starts at level 3 (caller of `__index`/`__newindex`'s caller, since the helper adds one frame). Each iteration:

- Returns `true` if the current frame is a method.
- Returns `false` if there's no frame at this level (stack ran out without finding a method).
- Continues walking otherwise.

### Behavior matrix

| caller pattern                                                | pre-fix       | post-fix     |
| ------------------------------------------------------------- | ------------- | ------------ |
| `obj:method()` reading `self.private.x`                        | allowed       | allowed      |
| Method calling `pcall(function() self.private.x = 1 end)`      | allowed       | allowed      |
| Method calling an anonymous local helper that touches private  | allowed       | allowed      |
| Global function `function steal() return obj.private.x end`    | denied        | denied       |
| Local function `local function steal() ... end`                | denied        | denied       |
| `(function() return obj.private.x end)()` from outside         | **allowed**   | denied       |
| `pcall(function() return obj.private.x end)` from outside      | **allowed**   | denied       |

The two rows that change are the ones the bypass relied on. Every other call pattern is unaffected.

### Caveat on threat model

This still doesn't make `private` cryptographically private. A determined consumer can attach a function as a method on any class to gain `namewhat='method'` and access anything they want. The privacy mechanism is a best-effort accidental-access barrier, not a sandbox. The fix just raises the bar from "trivially three keystrokes" to "you have to define a real method first," which is enough to catch ordinary mistakes and casual external probing.

### Internal compatibility

Every `self.private` access in ox_lib's own classes (`timer`, `dui`, `scaleform`, `selector`, plus consumers of `lib.class`) is a direct read or write inside a named method body. None go through anonymous helpers that would have relied on the `namewhat == ''` allowance. So tightening doesn't break ox_lib internals.

### DIFF
--- a/imports/class/shared.lua
+++ b/imports/class/shared.lua
@@ -85,20 +85,27 @@ function mixins.new(class, ...)
         private = table.clone(obj.private)
 
         table.wipe(obj.private)
+
+        local function calledFromMethod()
+            local level = 3
+            while true do
+                local di = getinfo(level, 'n')
+                if not di then return false end
+                if di.namewhat == 'method' then return true end
+                level += 1
+            end
+        end
+
         setmetatable(obj.private, {
             __metatable = 'private',
             __tostring = void,
             __index = function(self, index)
-                local di = getinfo(2, 'n')
-
-                if di.namewhat ~= 'method' and di.namewhat ~= '' then return end
+                if not calledFromMethod() then return end
 
                 return private[index]
             end,
             __newindex = function(self, index, value)
-                local di = getinfo(2, 'n')
-
-                if di.namewhat ~= 'method' and di.namewhat ~= '' then
+                if not calledFromMethod() then
                     error(("cannot set value of private field '%s'"):format(index), 2)
                 end
 


thelindat:
It doesn't really matter that much and, as stated, it's trivial to call a private field from a method on another class - and much more likely to do on accident. Private fields exist to prevent accidents, not for security.

Kenshiin13:
> It doesn't really matter that much and, as stated, it's trivial to call a private field from a method on another class - and much more likely to do on accident. Private fields exist to prevent accidents, not for security.

The patterns that are common by accident are `pcall(function() ... end)` and IIFEs, both of which currently slip through the namewhat == '' clause. If the gate's purpose is preventing accidents, those are exactly the cases worth catching.

thelindat:
This is FiveM, barely anybody has heard of pcall and nobody uses or knows what IIFEs are. Anybody determined to mess around with private fields will do so unless it's fully patched, but it's already a hacky solution anyway.

Kenshiin13:
> This is FiveM, barely anybody has heard of pcall and nobody uses or knows what IIFEs are. Anybody determined to mess around with private fields will do so unless it's fully patched, but it's already a hacky solution anyway.

I don't agree with this. The incompetence of the consumer shouldn't be an excuse; if the lib provides a solution to a problem, it should match its own contract regardless of how the caller wrote their access.

The fact that another loose end exists also isn't an argument against tightening this one. Otherwise why bother introducing private fields at all? If the design intent is best-effort accident prevention, then catching the cases reachable by accident (pcall, IIFE) and leaving only the case that requires deliberate intent (defining a method on another class) is a strict improvement. The two states are:

Today: gate fires on function steal() (rare in practice) but lets through pcall(function() ... end) (extremely common).

After this PR: gate fires on both shapes, plus IIFEs. The only remaining bypass is "define a method on another class," which is never accidental.

The only thing the fix newly blocks is direct writes from closures:


SetTimeout(1000, function() self.private.x = y end)
The encapsulated form still works:


SetTimeout(1000, function() self:setX(y) end)
So anything the fix breaks is the case where the consumer was reaching past the abstraction in the first place. That's the gate doing its job.

Happy to look at cross-class access in a follow-up PR if you want true-class-private semantics, but that's a bigger restructuring (probably moving away from the debug.getinfo check toward proxy methods or a class-method registry). It shouldn't block this one.

thelindat:
Because of the way debug.getinfo works, calling self:method() doesn't report itself as a method which is why anonymous calls were given an exception in the first place. https://github.com/overextended/ox_lib/issues/549

PR #781, opened 2026-05-17
allow keybinds while pause menu is active

Mesrine67:
Adds an optional `allowInPauseMenu` flag to `lib.addKeybind`.

By default, keybinds keep the existing behavior and are ignored while `IsPauseMenuActive()` is true. When `allowInPauseMenu` is set to true, the keybind can still trigger while the pause menu/frontend is active.

This is useful for resources that use frontend pause menu rendering, such as ped previews, while still needing their mapped keybinds to work.


thelindat:
This requires accompanying changes in the ts package and docs.

Mesrine67:
Docs PR: https://github.com/overextended/overextended.github.io/pull/277

PR #783, opened 2026-05-18
feat(hook): add rate limiting to HookPipeline

mysbryce:
### Summary
Adds built-in rate limiting support to `HookPipeline` via `setRateLimit()`. The rate limit is evaluated **before any hook runs**, so blocked dispatches have zero overhead on the hook chain.

---

### Changes
- Add `setRateLimit(options)` — configure max calls per time window per key
- Add `clearRateLimit()` — remove an existing rate limit from the pipeline
- Add `_checkRateLimit(payload)` — internal fixed-window counter
- Initialize `_rateLimit` and `_rateLimitState` fields in `constructor`

---

### API

```lua
myPipeline:setRateLimit({
    maxCalls   = 5,
    window     = 1000, -- ms

    -- optional: custom key (default: player source)
    key = function(payload)
        return tostring(payload.playerId or source)
    end,

    -- optional: called when limit is exceeded
    onExceeded = function(key, payload)
        print(('player %s exceeded rate limit'):format(key))
    end,
})

myPipeline:clearRateLimit()
```

---

### How it works

Uses a **fixed-window** strategy — each key gets a counter that resets after `window` ms.
If the counter exceeds `maxCalls`, `dispatch` returns `{ ok = false, size = 0 }` immediately without running any hooks.

```
dispatch(payload)
    │
    ▼
_checkRateLimit()      ← checked before any hook
    │
    ├─ exceeded → return { ok = false, size = 0 }
    │
    └─ allowed  → run hooks as normal
```

---

### Notes
- Works on both **server-side** and **client-side** pipelines
- Default key is `source` (player server ID), suitable for server-side events
- For client-side pipelines, a custom `key` function is recommended
- `key` returning `nil` or empty string falls back to `source` automatically

---

### Testing

- [x] Rate limit blocks dispatch after `maxCalls` within `window`
- [x] Counter resets correctly after the window expires
- [x] `onExceeded` callback is invoked on breach
- [x] `clearRateLimit()` removes the limit and resets state
- [x] Custom `key` function is respected
- [x] Fallback to `source` when `key` returns nil

### DIFF
--- a/imports/hook/shared.lua
+++ b/imports/hook/shared.lua
@@ -1,9 +1,10 @@
 --[[
-    https://github.com/overextended/ox_lib
+https://github.com/overextended/ox_lib
 
-    This file is licensed under LGPL-3.0 or higher <https://www.gnu.org/licenses/lgpl-3.0.en.html>
+This file is licensed under LGPL-3.0 or higher <https://www.gnu.org/licenses/lgpl-3.0.en.html>
 
-    Copyright © 2025 Linden <https://github.com/thelindat>
+Copyright © 2025 Linden <https://github.com/thelindat>
+
 ]]
 
 local hooks = {}
@@ -16,11 +17,13 @@ end)
 
 ---@class HookPipeline : OxClass
 ---@field private new HookPipelineConstructor
+
 lib.hook = lib.class('HookPipeline')
 
 ---@class HookPipelineConstructor
 ---@overload fun(self: HookPipeline, event: string, filter: fun(hook: table, payload: unknown): boolean): HookPipeline
 ---@private
+
 ---Creates a hook pipeline for a specific event.
 ---
 ---The pipeline manages a collection of registered hooks and controls execution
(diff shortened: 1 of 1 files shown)


thelindat:
Despite being fully written by an LLM, I’ll bite. Why?

mysbryce:
> Despite being fully written by an LLM, I’ll bite. Why?

Yeah, I used an LLM to help format the text and code neatly, but the feature comes from a practical issue.

The main idea is to prevent exploitation and spam abuse of critical pipelines (such as inventory or interaction hooks). Instead of repeating rate-limiting logic across multiple hooks, this provides a centralized, built-in way to instantly reject dispatches. Since it evaluates before the chain runs, it saves server resources when a player tries to flood triggers.

I don’t know the others, but in Thailand, we had this problem, and we used a lot of ox_lib.

thelindat:
These are local events; players cannot flood the server with requests.

fxserver also has configurable rate limits for net events.

PR #785, opened 2026-05-21
Custom xenon headlights color

xoomer:


thelindat:
[SetVehicleXenonLightsCustomColor](https://docs.fivem.net/natives/?_0x1683E7F0).

> This native is not synced between players.

This cannot be handled in vehicle properties.

PR #790, opened 2026-06-04
feat(radial) add ability to add & remove items from submenus

Gravxd:
This PR adds two new methods for the Radial Menu to allow better/more convenient management of sub menus for common scenarios such as adding & removing options within submenus based on certain conditions (e.g jobs, coords, states etc)

```lua
---Registers an item or array of items in a specific radial sub menu.
---@param string parentMenuId
---@param items RadialMenuItem | RadialMenuItem[]
function lib.addRadialSubItem(parentMenuId, items)
```

```lua
-- Removes an item from a radial submenu with the given parentMenuId
---@param parentMenuId string
---@param id string
function lib.removeRadialSubItem(parentMenuId, id)
```

When i asked around in discord and the docs etc, the current "way to do it" is re-registering the entire submenu e.g if i have an actions menu (globalmenu->actions) i'd have to re-register the entire menu and keep doing so based on conditions.

With these new methods, i can simply use the add and remove functions above to pluck options as needed.

Example usage based on the doc's [example menu](https://overextended.dev/ox_lib/Modules/Interface/Client/radial#usage-example)
```lua
lib.addRadialSubItem('police_menu', {
      {
          id = 'police.one',
          label = 'Hello',
          icon = 'fa-solid fa-car',
          onSelect = function()
              print('Hello')
          end,
      },
      {
          id = 'police.two',
          label = 'World',
          icon = 'fa-solid fa-print',
          onSelect = function()
              print('World')
          end,
      }
  })
```

And then if i wanted to remove a specific item based on my position, or whatever use case (this is similar to the use case for the docs example garage button being added/removed based on coords)

```lua
lib.removeRadialSubItem('police_menu', 'police.one')
```

Documentation PR was/is: https://github.com/overextended/overextended.github.io/pull/259


Gravxd:
Closed to move this to a separate branch with an additional bug fix.

PR #791, opened 2026-06-08
fix: stop mutating caller-supplied tables

Kenshiin13:
### Description

Five public functions in ox_lib write back into the config tables passed by the caller. None of this is documented. A caller that shares one config table across multiple calls sees the second call silently behave differently from the first because the first one mutated their table.

The offenders:

| function | what it writes back |
| --- | --- |
| `lib.marker.new` | `options.type` (normalized), `options.width`/`options.height` (coerced to float), and attaches a metatable to the input |
| `lib.notify` | nils `data.sound`, defaults `data.position` onto the caller's table |
| `lib.inputDialog` | rewrites legacy string entries in `rows` to `{ type = 'input', label = rows[i] }` |
| `lib.points.new` | writes `id`, `remove`, `coords` (replaced with vector), `distance`, `radius`, plus any args[3] fields onto the caller's data |
| `lib.zones.poly`/`box`/`sphere` | attaches `id`, `polygon`, `coords`, `__type`, `radius`, `remove`, `contains`, `setDebug`, etc. onto the caller's data |

### Reproduction

```lua
local notification = {
    title = 'Saved',
    description = 'OK',
    sound = { name = 'CHECKPOINT_PERFECT', set = 'HUD_MINI_GAME_SOUNDSET' },
}

lib.notify(notification)
-- notification.sound is now nil.

lib.notify(notification)
-- second call has no sound; the first wiped it.
```

### Fix

Each function now `table.clone`s its input at the top and mutates the clone. The caller's table is untouched after the call. `setZone` (the internal helper used by the three zone constructors) still mutates its argument, but it now always receives the clone, so no caller-visible mutation remains.

### DIFF
--- a/imports/marker/client.lua
+++ b/imports/marker/client.lua
@@ -101,11 +101,11 @@ end
 
 ---@param options MarkerProps
 function lib.marker.new(options)
-  options.type =
-    type(options.type) == 'string' and markerTypes[options.type]
-    or type(options.type) == 'number' and options.type or nil
+  local self = setmetatable(table.clone(options), marker_mt)
 
-  local self = setmetatable(options, marker_mt)
+  self.type =
+    type(self.type) == 'string' and markerTypes[self.type]
+    or type(self.type) == 'number' and self.type or nil
 
   self.width += .0
   self.height += .0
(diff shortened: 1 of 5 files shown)


thelindat:
There's a non-zero chance that people rely on the current behaviour of points and zones.

Kenshiin13:
> There's a non-zero chance that people rely on the current behaviour of points and zones.

Fair point, I'll leave those untouched

PR #794, opened 2026-06-08
fix(requestModel): add validation for nil model input

AnishBplayz:
This pull request improves the handling of invalid or missing model values when loading models in the client code. The main changes add explicit checks for `nil` models and provide clearer error handling to prevent runtime errors and improve debugging.

**Model validation and error handling:**

* Added a check in `lib.requestModel` to raise an error if the `model` parameter is `nil`, preventing attempts to load invalid models.
* Updated `loadPropModel` to return a failure and an error message if the `model` argument is `nil`, ensuring that invalid input is handled gracefully.

**Progress prop loading improvements:**

* Modified `startProgress` to only attempt to load prop models if the `model` value exists, both for single and multiple props, preventing unnecessary errors and improving robustness.

### DIFF
--- a/imports/requestModel/client.lua
+++ b/imports/requestModel/client.lua
@@ -11,6 +11,10 @@
 ---@param timeout number? Approximate milliseconds to wait for the model to load. Default is 10000.
 ---@return number model
 function lib.requestModel(model, timeout)
+    if not model then
+        error('attempted to load invalid model (nil)')
+    end
+
     if type(model) ~= 'number' then model = joaat(model) end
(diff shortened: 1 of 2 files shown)


unitysync:
not accepting ai slop

AnishBplayz:
Correction - The copilot review happens automatically, not my fault you are dumb to think it as AI slop...

unitysync:
> Correction - The copilot review happens automatically, not my fault you are dumb to think it as AI slop...

commit description and PR read like AI. its unnecessary since there are already guard clauses to prevent this sort of thing.

PR #796, opened 2026-06-12
feat: Add four collection types: lib.set, lib.ringbuffer, lib.lru, lib.heap

Kenshiin13:
## Add four collection types: `lib.set`, `lib.ringbuffer`, `lib.lru`, `lib.heap`

Each is its own `OxClass`, one file per type, no cross-dependencies. The PR is split into four feature commits plus one follow-up commit that applies maintainer feedback (varargs constructors, sparse-set storage for Set, `__pairs`/`__len` metamethods on the iterables).

## `lib.set`: Unique values with set operations

Use when you want membership semantics with real ops (union, intersection, difference, subset). Storage is dense-array + hashmap (sparse-set style), so `add`/`remove`/`has` are O(1) and the underlying array gives clean iteration. Swap-remove on `:remove`. Insertion order is preserved until the first delete.

```lua
local online = lib.set:new(1, 2, 3, 4)
local admins = lib.set:new(2, 5)

online:has(2)              -- true
online:add(6):add(7)       -- chainable
online:remove(1)           -- → true (was present)
online:size()              -- 5
#online                    -- 5  (__len)

-- set algebra
local both       = online:intersection(admins)   -- Set { 2 }
local everybody  = online:union(admins)          -- Set { 2, 3, 4, 5, 6, 7 }
local nonAdmins  = online:difference(admins)     -- Set { 3, 4, 6, 7 }

online:isSubsetOf(everybody)                     -- true
online:equals(lib.set:new(2, 3, 4, 6, 7))        -- true

-- iteration
for playerId in pairs(online) do                 -- __pairs (single-value yield)
    print(playerId)
end

for playerId in online:each() do                 -- or explicit
    print(playerId)
end

online:toArray()                                 -- returns Array, chain :map/:filter/:reduce
```

---

## `lib.ringbuffer`: Fixed-capacity rolling log

Use when you want "the last N of something." Kill feeds, chat backlogs, recent damage events, audit trails, recent errors. `push` returns whatever was evicted, so you can hook a cleanup pass without a separate query.

```lua
local recent = lib.ringbuffer:new(3)

recent:push('a')           -- nil  (room left)
recent:push('b')           -- nil
recent:push('c')           -- nil  (buffer full)
recent:push('d')           -- 'a'  (evicted)

recent:get(1)              -- 'b'  (oldest)
recent:get(-1)             -- 'd'  (newest; negative indexes from end)
recent:size()              -- 3
#recent                    -- 3   (__len)
recent:isFull()            -- true

recent:toArray()           -- Array { 'b', 'c', 'd' }  (oldest first)

for entry in pairs(recent) do     -- __pairs
    print(entry)
end
```

---

## `lib.lru`: Bounded cache with LRU eviction

Use when you want to cache things but bound the size. HTTP response memo, oxmysql query cache, player-metadata lookup, "remember the last N times someone did X." Unbounded tables leak memory in long-running servers; this gives you a hard ceiling and evicts the least-recently-used entry on overflow.

```lua
local cache = lib.lru:new(100)

cache:set('player:42', metadata)
cache:get('player:42')                -- bumps to most-recently-used
cache:has('player:42')                -- true (no bump)
cache:peek('player:42')               -- read without bumping

-- evictions are surfaced
local evictedKey, evictedValue = cache:set('player:101', otherMetadata)
if evictedKey then
    print(('cache evicted %s'):format(evictedKey))
end

cache:delete('player:42')             -- → true
cache:size()                          -- 1
#cache                                -- 1   (__len)
cache:capacity()                      -- 100
cache:clear()

-- iterate MRU → LRU
for k, v in pairs(cache) do           -- __pairs (two-value yield)
    print(k, v)
end
```

---

## `lib.heap`: Binary heap / priority queue

Use when you need the smallest (or comparator-first) item out of a growing set in O(log n). Dispatch by priority, scheduled events.

```lua
-- min-heap (default)
local h = lib.heap:new()
h:push(5):push(1):push(3):push(2)
h:peek()                              -- 1
h:pop()                               -- 1
h:pop()                               -- 2
h:size()                              -- 2
#h                                    -- 2   (__len)

-- varargs push
local h2 = lib.heap:new()
h2:push(5, 1, 3, 2, 8, 4)
h2:drain()                            -- Array { 1, 2, 3, 4, 5, 8 }

-- max-heap via comparator
local maxHeap = lib.heap:new(function(a, b) return a > b end)
maxHeap:push(3, 1, 4, 1, 5, 9)
maxHeap:pop()                         -- 9

-- priority queue of objects
local dispatch = lib.heap:new(function(a, b) return a.priority < b.priority end)
dispatch:push(
    { priority = 5, call = 'noise complaint' },
    { priority = 1, call = 'shots fired' },
    { priority = 3, call = 'traffic stop' }
)

while not dispatch:isEmpty() do
    local job = dispatch:pop()
    print(job.priority, job.call)
end
-- 1  shots fired
-- 3  traffic stop
-- 5  noise complaint
```

### DIFF
--- a/imports/heap/shared.lua
+++ b/imports/heap/shared.lua
@@ -8,7 +8,7 @@
 
 ---@class Heap : OxClass
 ---@field private new HeapConstructor
-local Heap = lib.class('Heap')
+lib.heap = lib.class('Heap')
 
 local function defaultLess(a, b) return a < b end
 
@@ -43,7 +43,7 @@ end
 ---@overload fun(self: Heap, comparator?: fun(a: any, b: any): boolean): Heap
 ---@private
 ---@param comparator? fun(a: any, b: any): boolean Returns true when `a` should come before `b`. Defaults to `a < b` (min-heap).
-function Heap:constructor(comparator)
+function lib.heap:constructor(comparator)
     if comparator ~= nil and type(comparator) ~= 'function' then
         error("comparator must be a function or nil", 2)
     end
@@ -54,7 +54,7 @@ end
 
 ---@param ... any
 ---@return Heap self
-function Heap:push(...)
+function lib.heap:push(...)
     local items = self.private.items
     local lt = self.private.lt
     local n = select('#', ...)
@@ -68,7 +68,7 @@ function Heap:push(...)
 end
 
 ---@return any?
-function Heap:pop()
+function lib.heap:pop()
     local items = self.private.items
     local n = #items
     if n == 0 then return nil end
@@ -87,28 +87,28 @@ function Heap:pop()
 end
 
 ---@return any?
-function Heap:peek()
+function lib.heap:peek()
     return self.private.items[1]
 end
 
 ---@return integer
-function Heap:size()
+function lib.heap:size()
     return #self.private.items
 end
 
 ---@return boolean
-function Heap:isEmpty()
+function lib.heap:isEmpty()
     return #self.private.items == 0
 end
 
 ---@return Heap self
-function Heap:clear()
+function lib.heap:clear()
     self.private.items = {}
     return self
 end
 
 ---@return Array sorted Newly-allocated array drained in pop order.
-function Heap:drain()
+function lib.heap:drain()
     local out = lib.array:new()
     local n = 0
     while #self.private.items > 0 do
@@ -118,7 +118,6 @@ function Heap:drain()
     return out
 end
 
-Heap.__len = Heap.size
+lib.heap.__len = lib.heap.size
 
-lib.heap = Heap
 return lib.heap
--- a/imports/heap/shared.lua
+++ b/imports/heap/shared.lua
@@ -52,12 +52,18 @@ function Heap:constructor(comparator)
     self.private.lt = comparator or defaultLess
 end
 
----@param value any
+---@param ... any
 ---@return Heap self
-function Heap:push(value)
+function Heap:push(...)
     local items = self.private.items
-    items[#items + 1] = value
-    siftUp(items, #items, self.private.lt)
+    local lt = self.private.lt
+    local n = select('#', ...)
+
+    for i = 1, n do
+        items[#items + 1] = (select(i, ...))
+        siftUp(items, #items, lt)
+    end
+
     return self
 end
 
@@ -101,20 +107,6 @@ function Heap:clear()
     return self
 end
 
----@param values any[]
----@return Heap self
-function Heap:pushAll(values)
-    local items = self.private.items
-    local lt = self.private.lt
-
-    for i = 1, #values do
-        items[#items + 1] = values[i]
-        siftUp(items, #items, lt)
-    end
-
-    return self
-end
-
 ---@return Array sorted Newly-allocated array drained in pop order.
 function Heap:drain()
     local out = lib.array:new()
@@ -125,6 +117,8 @@ function Heap:drain()
     end
     return out
 end
+
+Heap.__len = Heap.size
 
 lib.heap = Heap
 return lib.heap
--- a/imports/heap/shared.lua
+++ b/imports/heap/shared.lua
@@ -111,7 +111,7 @@ function lib.heap:clear()
     return self
 end
 
----@return Array sorted newly-allocated array drained in pop order.
+---@return Array sorted Newly-allocated array drained in pop order.
 function lib.heap:drain()
     local out = lib.array:new()
     local n = 0
--- a/imports/heap/shared.lua
+++ b/imports/heap/shared.lua
@@ -1,0 +1,130 @@
+--[[
+    https://github.com/overextended/ox_lib
+
+    This file is licensed under LGPL-3.0 or higher <https://www.gnu.org/licenses/lgpl-3.0.en.html>
+
+    Copyright © 2025 Linden <https://github.com/thelindat>
+]]
+
+---@class Heap : OxClass
+---@field private new HeapConstructor
+local Heap = lib.class('Heap')
+
+local function defaultLess(a, b) return a < b end
+
+local function siftUp(items, i, lt)
+    while i > 1 do
+        local parent = i >> 1
+        if not lt(items[i], items[parent]) then return end
+        items[i], items[parent] = items[parent], items[i]
+        i = parent
+    end
+end
+
+local function siftDown(items, i, n, lt)
+    while true do
+        local left = i * 2
+        if left > n then return end
+
+        local smallest = i
+        if lt(items[left], items[smallest]) then smallest = left end
+
+        local right = left + 1
+        if right <= n and lt(items[right], items[smallest]) then smallest = right end
+
+        if smallest == i then return end
+
+        items[i], items[smallest] = items[smallest], items[i]
+        i = smallest
+    end
+end
+
+---@class HeapConstructor
+---@overload fun(self: Heap, comparator?: fun(a: any, b: any): boolean): Heap
+---@private
+---@param comparator? fun(a: any, b: any): boolean Returns true when `a` should come before `b`. Defaults to `a < b` (min-heap).
+function Heap:constructor(comparator)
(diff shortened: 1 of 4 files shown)


thelindat:
Could you use varargs rather than tables for passing values to constructors, pushAll, etc.
Use an ordered set (an array and a hashmap) similar to JS sets since that also covers sparse sets. You could add some metamethods for convenience here too, at the very least __iter allows a generic for in loop.

Kenshiin13:
>You could add some metamethods for convenience here too, at the very least __iter allows a generic for in loop.

Is `__iter` metamethod a thing in CFX Lua 5.4 implementation? Doesn't seem to honor it. I added `__pairs`.

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+    https://github.com/overextended/ox_lib
+
+    This file is licensed under LGPL-3.0 or higher <https://www.gnu.org/licenses/lgpl-3.0.en.html>
+
+    Copyright © 2025 Linden <https://github.com/thelindat>
+]]
+
+---@class Heap : OxClass
+---@field private new HeapConstructor
+local Heap = lib.class('Heap')
```
thelindat: Assigning a local like this and then assigning it to the lib global causes issues with LLS - take note of other classes which were all changes to direct assignments.

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+
+---@param ... any
+---@return Heap self
+function Heap:push(...)
+    local items = self.private.items
+    local lt = self.private.lt
+    local n = select('#', ...)
+
+    for i = 1, n do
+        items[#items + 1] = (select(i, ...))
```
thelindat: Probably more efficient to create a table `{...}` and then use `table.move` to combine it with the heap.

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+
+        local right = left + 1
+        if right <= n and lt(items[right], items[smallest]) then smallest = right end
+
+        if smallest == i then return end
+
+        items[i], items[smallest] = items[smallest], items[i]
+        i = smallest
+    end
+end
```
thelindat, line 40: Wouldn't `table.sort` be the best approach here? At least depending on how many elements there are, anyway.

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+
+---@return boolean
+function Heap:isEmpty()
+    return #self.private.items == 0
+end
+
+---@return Heap self
+function Heap:clear()
+    self.private.items = {}
```
thelindat: `table.wipe`

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+    return self
+end
+
+---@return Array sorted Newly-allocated array drained in pop order.
```
thelindat: `N`ewly -> `n`ewly?

imports/heap/shared.lua
```diff
@@ -0,0 +1,124 @@
...
+function Heap:drain()
+    local out = lib.array:new()
+    local n = 0
+    while #self.private.items > 0 do
+        n += 1
+        out[n] = self:pop()
+    end
+    return out
+end
```
thelindat: Wouldn't this be incredibly slow as it constantly shifts all elements as it drains?

imports/lru/shared.lua
```diff
@@ -0,0 +1,151 @@
...
+end
+
+---@return integer
+function Lru:capacity()
+    return self.private.capacity
+end
+
+---@return Lru self
+function Lru:clear()
+    self.private.map = {}
```
thelindat: `table.wipe`

imports/set/shared.lua
```diff
@@ -0,0 +1,156 @@
...
+---@private
+function Set:constructor(...)
+    self.private.items = {}
+    self.private.index = {}
+
+    local n = select('#', ...)
+    for i = 1, n do
+        self:add((select(i, ...)))
+    end
+end
```
thelindat, line 24: Renaming index to map is a little more intuitive imo.

imports/set/shared.lua
```diff
@@ -0,0 +1,156 @@
...
+    if i ~= lastI then
+        local last = items[lastI]
+        items[i] = last
+        self.private.index[last] = i
+    end
+
+    items[lastI] = nil
+    self.private.index[value] = nil
+    return true
+end
```
thelindat, line 56: Use `table.move` to shift the set down, I want to keep the set ordered by insertion.

imports/set/shared.lua
```diff
@@ -0,0 +1,156 @@
...
+
+---@return boolean
+function Set:isEmpty()
+    return #self.private.items == 0
+end
+
+---@return Set self
+function Set:clear()
+    self.private.items = {}
+    self.private.index = {}
```
thelindat: `table.wipe`

imports/set/shared.lua
```diff
@@ -0,0 +1,156 @@
...
+    local i = 0
+    return function()
+        i += 1
+        return items[i]
+    end
+end
+
+function Set:__pairs()
+    return self:each(), nil, nil
+end
```
thelindat:
```lua
function Set:__pairs()
    return next, self.private.items
end

Set.__ipairs = Set.__pairs
```

thelindat (changes requested):

imports/heap/shared.lua
Kenshiin13: Fixed. To be fair, [`imports/selector/shared.lua`](https://github.com/overextended/ox_lib/blob/39e8cce0bbe5c2e76deaf536687ad133465642cd/imports/selector/shared.lua#L177) uses the same shape. Worth addressing in a followup PR/commit.

imports/heap/shared.lua
Kenshiin13: You're right. Fixed.

imports/heap/shared.lua
Kenshiin13: `siftDown` only touches the path from the swapped root down to its final spot. For a heap of `n` items that's about `log n` comparisons and at most `log n` swaps per pop. `table.sort` would re-sort the whole array on every pop, which is `n log n`.

The whole point of using a heap instead of just keeping a sorted array is that you never pay the full sort cost.

imports/heap/shared.lua
Kenshiin13: Fixed.

imports/lru/shared.lua
Kenshiin13: Fixed.

imports/set/shared.lua
Kenshiin13: Fixed.

imports/heap/shared.lua
Kenshiin13: Fixed.

imports/set/shared.lua
Kenshiin13: `index` documents that the field stores positions in items rather than a generic key-value mapping (which is what map means in `lib.lru`). But it's a one-word rename and your call on the codebase style. Want me to swap to map?

imports/heap/shared.lua
Kenshiin13: No shift, `pop` swaps the last element into the root and sifts down. `items[1] = items[n]`, `items[n] = nil`, `siftDown`. All constant-time except the `siftDown` which is `log n`. Drain calls pop `n` times so total cost is `n log n`, same as table.sort.

imports/set/shared.lua
Kenshiin13: Switched to table.move + index rewrite for the shifted range. Order preserved on remove now. Remove goes from `O(1)` to `O(n)` but that's the cost of the guarantee.

imports/set/shared.lua
Kenshiin13: Fixed.

PR #797, opened 2026-06-13
feat(radial) Add new methods to add & remove items from submenus

Gravxd:
Opened a second PR since i wanted to move this to a new branch as well as had to fix something on it.

This PR adds two new methods for the Radial Menu to allow better/more convenient management of sub menus for common scenarios such as adding & removing options within submenus based on certain conditions (e.g jobs, coords, states etc)

This makes using the radial menu features a lot more convenient and easier when trying to dynamically update what options are available to a player based on the conditions I listed above. 

The current approach is that you have to re-register the whole item/menu each time which is inefficient.

```lua
---Registers an item or array of items in a specific radial sub menu.
---@param parentMenuId string
---@param items RadialMenuItem | RadialMenuItem[]
function lib.addRadialSubItem(parentMenuId, items)
```

```lua
-- Removes an item from a radial submenu with the given parentMenuId
---@param parentMenuId string
---@param id string
function lib.removeRadialSubItem(parentMenuId, id)
```

When i asked around in discord and the docs etc, the current "way to do it" is re-registering the entire submenu e.g if i have an actions menu (globalmenu->actions) i'd have to re-register the entire menu and keep doing so based on conditions.

With these new methods, i can simply use the add and remove functions above to pluck options as needed.

Example usage based on the doc's [example menu](https://overextended.dev/ox_lib/Modules/Interface/Client/radial#usage-example)
```lua
lib.addRadialSubItem('police_menu', {
      {
          id = 'police.one',
          label = 'Hello',
          icon = 'fa-solid fa-car',
          onSelect = function()
              print('Hello')
          end,
      },
      {
          id = 'police.two',
          label = 'World',
          icon = 'fa-solid fa-print',
          onSelect = function()
              print('World')
          end,
      }
  })
```

An additional example would be a generic radial submenu with options e.g for player interactions and then based on either my permissions or a unique condition, I can make more options available or take some away.

And then if i wanted to remove a specific item based on my position, or whatever use case (this is similar to the use case for the docs example garage button being added/removed based on coords)

```lua
lib.removeRadialSubItem('police_menu', 'police.one')
```

Documentation PR was/is: https://github.com/overextended/overextended.github.io/pull/259

### DIFF
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -194,6 +194,64 @@ function lib.removeRadialItem(id)
     refreshRadial(id)
 end
 
+---Registers an item or array of items in a specific radial sub menu.
+---@param string parentMenuId
+---@param items RadialMenuItem | RadialMenuItem[]
+function lib.addRadialSubItem(parentMenuId, items)
+
+    local parentMenu = menus[parentMenuId]
+    if not parentMenu then
+        return error('No radial menu with such id found.')
+    end
+
+    local invokingResource = GetInvokingResource()
+
+    items = table.type(items) == 'array' and items or { items }
+
+    for i = 1, #items do
+        local item = items[i]
+        item.resource = invokingResource
+
+        for j = 1, #parentMenu.items do
+            if parentMenu.items[j].id == item.id then
+                parentMenu.items[j] = item
+            end
+
+            if j == #parentMenu.items then
+                parentMenu.items[#parentMenu.items + 1] = item
+            end
+        end
+        
+    end
+
+    if isOpen then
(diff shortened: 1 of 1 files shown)


thelindat:
Rather than creating new functions, why not add a second argument to addRadialItem/removeRadialItem that defaults to global?

Gravxd:
> Rather than creating new functions, why not add a second argument to addRadialItem/removeRadialItem that defaults to global?

Yeh can do if that's preferred and thinking about it probably makes more sense - i'll edit the PR and relevant PR on the docs repo

Gravxd:
If the new commit is fine and is to be merged, i'll amend the docs PR

PR #798, opened 2026-06-14
feat: add insertion-ordered Map & refactor Set to use it

Kenshiin13:
## Add `lib.map`, refactor `lib.set` to use it

Adds an insertion-ordered key/value collection with amortized O(1) delete, and rewires `lib.set` to use it internally so Set's `remove` goes from O(n) to amortized O(1) without any public-API change.

## `lib.map`: insertion-ordered key/value collection

Storage is a dense entries array (in insertion order) plus a hashmap of `key → array position`. Deletes mark the slot with a sentinel rather than shifting; a periodic in-place rebuild compacts when tombstoned slots outnumber live ones.

```lua
local m = lib.map:new({ a = 1, b = 2 })   -- optional initial table
m:set('c', 3):set('d', 4)                 -- chainable; insertion-ordered
m:get('a')                                -- 1
m:has('a')                                -- true
m:delete('b')                             -- true; amortized O(1), no shift
m:size()                                  -- 3
#m                                        -- 3   (__len)

for k, v in pairs(m) do                   -- __pairs; insertion order
    print(k, v)
end

for k in m:keys() do ... end              -- iterator over keys
for v in m:values() do ... end            -- iterator over values
for k, v in m:entries() do ... end        -- iterator over (k, v); :each is an alias

m:toTable()                               -- plain Lua table snapshot, for json.encode / interop
m:toArray()                               -- Array of { key, value } in insertion order
```

### Notes

- **Constructor accepts another Map**. Passing one preserves insertion order (because `pairs` on a Map honors `__pairs`).
- **`set(k, nil)` is equivalent to `delete(k)`**. Mirrors Lua table semantics.
- **Updating an existing key keeps its position**
- **Iteration order from a Lua table literal is not guaranteed** by Lua. If order matters at construction time, build the Map explicitly with chained `:set` calls.

## `lib.set` refactor

`lib.set` now wraps a `lib.map` internally (value of every entry is `true`). All public methods and metamethods unchanged: same constructor, same `add`/`remove`/`has`/`each`/`__pairs`/`__ipairs`/`__len`/set-algebra surface, same `(i, v)` yield shape.

Behavioral diff:

- `remove` went from **O(n)** (shift + reindex) to **amortized O(1)** via Map's tombstone delete.
- Insertion order still preserved.
- No API change for callers.

### DIFF
--- a/imports/heap/shared.lua
+++ b/imports/heap/shared.lua
@@ -6,30 +6,40 @@
     Copyright © 2025 Linden <https://github.com/thelindat>
 ]]
 
+---@alias HeapComparator fun(a: any, b: any): boolean
+
 ---@class Heap : OxClass
 ---@field private new HeapConstructor
 lib.heap = lib.class('Heap')
 
+---@type HeapComparator
 local function defaultLess(a, b) return a < b end
 
+---@param items any[]
+---@param i integer
+---@param lt HeapComparator
 local function siftUp(items, i, lt)
     while i > 1 do
-        local parent = i >> 1
+        local parent = i >> 1 ---@type integer
         if not lt(items[i], items[parent]) then return end
         items[i], items[parent] = items[parent], items[i]
         i = parent
     end
 end
 
+---@param items any[]
+---@param i integer
+---@param n integer
+---@param lt HeapComparator
 local function siftDown(items, i, n, lt)
     while true do
-        local left = i * 2
+        local left = i * 2 ---@type integer
         if left > n then return end
 
-        local smallest = i
+        local smallest = i ---@type integer
         if lt(items[left], items[smallest]) then smallest = left end
 
-        local right = left + 1
+        local right = left + 1 ---@type integer
         if right <= n and lt(items[right], items[smallest]) then smallest = right end
 
         if smallest == i then return end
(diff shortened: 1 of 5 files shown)


imports/heap/shared.lua
```diff
@@ -6,30 +6,40 @@
...
+---@type HeapComparator
 local function defaultLess(a, b) return a < b end
 
+---@param items any[]
+---@param i integer
+---@param lt HeapComparator
 local function siftUp(items, i, lt)
     while i > 1 do
-        local parent = i >> 1
+        local parent = i >> 1 ---@type integer
```
thelindat: Is there a particular reason to explicitly define the type when it's inferred from `i` already? Here and in a dozen other places.

imports/heap/shared.lua
```diff
@@ -111,10 +121,10 @@ function lib.heap:clear()
     return self
 end
 
----@return Array sorted Newly-allocated array drained in pop order.
+---@return Array sorted newly-allocated array drained in pop order.
```
thelindat: 🤯

imports/lru/shared.lua
```diff
@@ -16,14 +18,19 @@
 ---@field private new LruConstructor
 lib.lru = lib.class('Lru')
 
+---@param self Lru
+---@param node LruNode
 local function detach(self, node)
-    local prev, nxt = node.prev, node.next
+    local prev = node.prev ---@type LruNode?
+    local nxt = node.next ---@type LruNode?
```
thelindat: Rather than constantly setting the type, please ensure the object itself is properly typed.

thelindat (changes requested):
Get rid of all these type comments and properly annotate the structures.

imports/map/shared.lua
```diff
@@ -0,0 +1,231 @@
...
+---@return Map self
+function lib.map:set(key, value)
+    if key == nil then
+        error("map key cannot be nil", 2)
+    end
+
+    if value == nil then
+        self:delete(key)
+        return self
+    end
```
thelindat: While JS maps do technically allow you to pass null, undefined, or nothing, it does not actually delete the key from the map.

Personally I think `value` should be any non-nil value.

PR #805, opened 2026-06-20
feat(require): support ./ and ../ relative paths

Kenshiin13:
## Summary

Adds compatibility for relative module paths in `lib.require`, `lib.load`, and `lib.loadJson`, resolved against the file containing the call.

## What this PR does

| Caller file | Call | Resolves to |
|---|---|---|
| `<res>/server/test.lua` | `lib.require('./main')` | `<res>/server/main.lua` |
| `<res>/server/test.lua` | `lib.require('./test.lua')` | `<res>/server/test.lua` |
| `<res>/server/test.lua` | `lib.require('./util.helpers')` | `<res>/server/util/helpers.lua` |
| `<res>/server/sub/x.lua` | `lib.require('../main')` | `<res>/server/main.lua` |
| `<res>/server/sub/x.lua` | `lib.require('../../config')` | `<res>/config.lua` |
| `<res>/main.lua` | `lib.require('./foo')` | `<res>/foo.lua` |
| `<res>/server/x.lua` | `lib.loadJson('./shops')` | reads `<res>/server/shops.json` |
| `<res>/server/x.lua` | `lib.loadJson('./shops.json')` | `.json` extension is stripped |

## Notes

- **Cache key converges across forms.** `./foo` from `server/test.lua` and `server.foo` from anywhere both resolve to the same internal name, so they share the cached module.
- **`.lua` and `.json` extensions are stripped** from the relative segment so user-typed extensions Just Work. Absolute calls still expect no extension (pre-existing behavior, unchanged).
- **`@resource/...` cross-resource paths are unaffected.** The guard only matches `./` or `../` prefixes.
- **Non-relative calls take the early-return branch** at the top of `resolveRelative`, so existing call sites are identical to upstream.

## Behavior change worth flagging

Anyone who was writing `lib.require('./foo')` in upstream and relying on it loading the root-level `<resource>/foo.lua` (the accidental upstream behavior) will see the call now resolve to `<callerDir>/foo.lua` instead. The explicit-intent fix for `./` is a real semantic change in that case. Likely never written intentionally.

### DIFF
--- a/imports/require/shared.lua
+++ b/imports/require/shared.lua
@@ -21,6 +21,66 @@ package = {
 
 ---@param modName string
 ---@return string
+local function resolveRelative(modName)
+    if not modName:find('^%.%.?/') then return modName end
+
+    local src
+    local i = 3
+
+    while true do
+        local info = debug.getinfo(i, 'S')
+        if not info then break end
+
+        local source = info.source
+
+        if source and source:sub(1, 1) == '@' and not source:find('^@+ox_lib/imports/require/') then
+            src = source
+            break
+        end
+
+        i += 1
+    end
+
+    if not src then
+        error(("cannot resolve relative path '%s' without a file source"):format(modName), 4)
+    end
+
+    local dir = src:match('^@+[^/]+/(.+)/[^/]+%.lua$') or ''
+    local parts = {}
+
+    for part in dir:gmatch('[^/]+') do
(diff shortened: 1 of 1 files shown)


thelindat:
The require function is intended to function as closely to the standard lua_require implementation as possible, which means no relative paths.

PR #808, opened 2026-06-22
fix(progress): ensure prop data is not empty before processing

SnepCnep:
### Fix Empty `prop` Handling

Prevents errors when `data.prop` is an empty table (`prop = []`) by ensuring the prop loading function is only called when a valid prop definition exists.

**What Changed**

* Added `next(data.prop) ~= nil` to the prop validation

* Prevents `loadPropModel(model)` from being called without a model.
* Fixes errors caused by empty `prop` tables (`prop = []`).

### DIFF
--- a/resource/interface/client/progress.lua
+++ b/resource/interface/client/progress.lua
@@ -195,7 +195,7 @@ local function startProgress(data, nuiMessage)
         Wait(0)
     end
 
-    if data.prop then
+    if data.prop and next(data.prop) ~= nil then
         TriggerServerEvent('ox_lib:progressProps', nil)
     end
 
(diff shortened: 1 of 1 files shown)


thelindat:
Why are people sending empty tables and if people send invalid data then yeah, an error would happen.

SnepCnep:
> Why are people sending empty tables and if people send invalid data then yeah, an error would happen.

In some escrow resource there is are some so that is why.

thelindat:
I mean it expects a specific shape and they don’t send it. If I have to add a nil check here then might as well go over the entire payload to verify each and every property matches the expected types. Which, sure, is technically best practice but also not really my problem so long as invalid data isn’t exploitable and game breaking (ie crashes).

I would just raise errors in the case of invalid property types which leads us back to square one anyway.

PR #809, opened 2026-06-25
feat(string): add string utility functions

GitScorz:
Add string utility functions for capitalization, prefix/suffix checks, and pattern replacement

### DIFF
--- a/imports/string/shared.lua
+++ b/imports/string/shared.lua
@@ -63,4 +63,72 @@ function string.random(pattern, length)
     return table.concat(arr)
 end
 
+---Capitalizes the first lowercase letter of a string
+---@param str string
+---@return string
+function string.capitalize(str)
+    return str:gsub('^%l', string.upper)
+end
+
+---Checks if a string starts with a given prefix
+---@param str string
+---@param prefix string
+---@return boolean
+function string.startsWith(str, prefix)
+    return str:sub(1, #prefix) == prefix
+end
+
+---Checks if a string ends with a given suffix
+---@param str string
+---@param suffix string
+---@return boolean
+function string.endsWith(str, suffix)
+    return suffix == '' or str:sub(-#suffix) == suffix
+end
+
+---Checks if a string contains a given value
+---@param str string
+---@param value string
+---@param plain? boolean If true or nil, searches as plain text. If false, searches as Lua pattern.
+---@return boolean
+function string.contains(str, value, plain)
+    return str:find(value, 1, plain ~= false) ~= nil
+end
+
+---Checks if a string is empty
+---@param str string
+---@return boolean
+function string.isEmpty(str)
+    return str == ''
+end
+
+---Checks if a string is blank (empty or only whitespace)
+---@param str string
+---@return boolean
+function string.isBlank(str)
+    return str:match('^%s*$') ~= nil
+end
+
+---Escapes Lua pattern characters in a string
+---@param str string
+---@return string
+function string.escapePattern(str)
+    return str:gsub('([^%w])', '%%%1')
+end
+
+---Replaces all text or pattern matches in a string
+---@param str string
+---@param old string
+---@param new string
+---@param plain? boolean If true or nil, replaces plain text. If false, replaces Lua pattern matches.
+---@return string
+function string.replace(str, old, new, plain)
+    if plain ~= false then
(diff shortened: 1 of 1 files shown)


thelindat:
None of these seem particularly useful imho

GitScorz:
The main benefit I had in mind was readability/intent at call sites rather than new functionality, e.g. str:contains(value) or str:startsWith(prefix) being clearer than repeated find/sub usage

GitScorz:
And also faster code writing instead of repeated small helper implementations across resources..

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+---@param suffix string
+---@return boolean
+function string.endsWith(str, suffix)
+    return suffix == '' or str:sub(-#suffix) == suffix
+end
+
+---Checks if a string contains a given value
+---@param str string
+---@param value string
+---@param plain? boolean If true or nil, searches as plain text. If false, searches as Lua pattern.
```
thelindat:
```lua
---@param plain? boolean Whether `value` is treated as a string literal or Lua pattern. Defaults to true.
```

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+---@return string
+function string.escapePattern(str)
+    return str:gsub('([^%w])', '%%%1')
+end
+
+---Replaces all text or pattern matches in a string
+---@param str string
+---@param old string
+---@param new string
+---@param plain? boolean If true or nil, replaces plain text. If false, replaces Lua pattern matches.
```
thelindat: Same comment change as contains.

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+---@return string
+function string.replace(str, old, new, plain)
```
thelindat: Use more descriptive parameter names, like `search` and `replacement`.

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+    if plain ~= false then
+        old = string.escapePattern(old)
+        new = new:gsub('%%', '%%%%')
+    end
+
+    return str:gsub(old, new)
```
thelindat: Wrap in parenthesis to only return the first value of gsub.

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+---@return boolean
+function string.isBlank(str)
+    return str:match('^%s*$') ~= nil
+end
+
+---Escapes Lua pattern characters in a string
+---@param str string
+---@return string
+function string.escapePattern(str)
+    return str:gsub('([^%w])', '%%%1')
```
thelindat: Tighten the pattern so it explicitly only escapes valid magic characters.

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+---@return boolean
+function string.contains(str, value, plain)
+    return str:find(value, 1, plain ~= false) ~= nil
+end
+
+---Checks if a string is empty
+---@param str string
+---@return boolean
+function string.isEmpty(str)
+    return str == ''
```
thelindat: I don't much care for this method 🤷

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
     return table.concat(arr)
 end
 
+---Capitalizes the first lowercase letter of a string
```
thelindat: The wording here suggests the string `"Hello"` would become `"HEllo"`, but the function specifically targets the first character of a string and converts it to uppercase if it is a lowercase letter.

imports/string/shared.lua
```diff
@@ -63,4 +63,72 @@ function string.random(pattern, length)
...
+end
+
+---Checks if a string is empty
+---@param str string
+---@return boolean
+function string.isEmpty(str)
+    return str == ''
+end
+
+---Checks if a string is blank (empty or only whitespace)
```
thelindat: `---Checks if a string is empty or contains only whitespace.`

thelindat (changes requested):

PR #811, opened 2026-07-01
fix(cron): don't push same-day cron runs to the next day

SlashyIsTaken:
### Problem

A cron whose time-of-day is still ahead today gets scheduled for the *next* day. For example `0 15 * * *` registered at 06:00 schedules tomorrow 15:00 instead of today 15:00. It happens for any `MM HH * * *` where the minute field is ≤ the current minute at registration time (so every `MM` = `00` cron started at `HH:00`), and it slips two days when the hour has also passed.

Steps to reproduce:
1. `lib.cron.new('0 15 * * *', job)` while the clock reads e.g. 06:00.
2. `task:getTimeAsString()` / the debug log shows the next run as *tomorrow* 15:00 rather than today 15:00.

### Cause

`getTimeUnit` computes the minute relative to the current minute, returning `value + maxUnits.min` when the minute has already passed this hour. In `OxTask:getNextTime` that rollover is carried into `day` whenever `self.hour` is set:

```lua
if minute >= maxUnits.min then
    local carry = math.floor(minute / maxUnits.min)
    minute = minute % maxUnits.min
    
    if not self.hour then
        hour += carry
    elseif not self.day then 
        day += carry
    ...
```

But when the target hour is later today (or wrapped into a following day via the hour >= maxUnits.hour block) that hour starts fresh and keeps its own minute, so the carry is spurious and advances the date.

### Fix

Only propagate the minute rollover when the resolved hour is still the current hour. Legitimate cases are unchanged: a wildcard hour (getTimeUnit returns the current hour, so the carry still advances it) and a pinned hour whose minutes are exhausted (carry rolls to the next day, as before).

No public API change, so no documentation update required.

### DIFF
--- a/imports/cron/server.lua
+++ b/imports/cron/server.lua
@@ -277,14 +277,16 @@ function OxTask:getNextTime()
         local carry = math.floor(minute / maxUnits.min)
         minute = minute % maxUnits.min
 
-        if not self.hour then
-            hour += carry
-        elseif not self.day then
-            day += carry
-        elseif not self.month then
-            month += carry
-        else
-            month += 12 * carry
+        if hour == currentDate.hour then
+            if not self.hour then
+                hour += carry
+            elseif not self.day then
+                day += carry
+            elseif not self.month then
+                month += carry
+            else
+                month += 12 * carry
+            end
         end
     end
 


Kenshiin13 (approved):
LGTM

FjamZoo (approved):

PR #812, opened 2026-07-10
feat(radial): Allow open specific radial menu as root

XCal-Dev:
### Description

I hope I wasn't blind and missed something obvious.

Adds lib.showRadialMenu(id), allowing a menu registered with lib.registerRadial to be opened directly as an independent root menu.

Previously, registered radial menus could only be reached by referencing them as a submenu from another radial item. This made context-specific radial menus depend on the global radial menu, even when they were intended to operate separately.

This allows resources to open their own registered radial menus without adding temporary items to, replacing, or otherwise modifying the global radial menu.

**Example**
```
lib.registerRadial({
    id = 'vehicle_actions',
    items = {
        {
            label = 'Toggle Engine',
            icon = 'power-off',
            onSelect = function()
                -- Toggle the vehicle engine.
            end
        },
        {
            label = 'More Options',
            icon = 'ellipsis',
            menu = 'vehicle_more_options'
        }
    }
})
lib.showRadialMenu('vehicle_actions')
```

### Behaviour

**A menu opened with lib.showRadialMenu:**
- Is displayed as a root menu without a back button.
- Does not affect the items in the global radial menu.
- Can open registered child menus normally.
- Shows the back button inside child menus and returns to the directly opened root.
- Transitions directly when called with a different registered root menu while one is already open.
- Transitions to the global radial menu when the normal radial keybind is pressed.
- Uses the existing radial focus, cursor, pause-menu, disabled-state, and control-blocking behaviour.

### Implementation

A separate rootRadial reference is used to distinguish the active root from its child menus.

This allows the existing showRadial transition and history logic to be reused while determining whether the current registered menu should be rendered as a root or submenu.

**Root state and menu history are reset when:**
- The radial menu is closed.
- Another registered root is opened.
- The user transitions back to the global radial menu.
- The global radial menu is opened normally.

### Compatibility

This does not change the existing lib.registerRadial, lib.addRadialItem, or global radial menu APIs. 
I have tested on the following.
1. latest release QBX with qbx_radialmenu with no issues or compatibility issues.
2. latest release QBX with my own custom radialmenu with no issues or compatibility issues.
3. All regular Radial functions work as intended, refreshRadial() still functions with lib.showRadialMenu()

### DIFF
--- a/package/client/interface/radial.ts
+++ b/package/client/interface/radial.ts
@@ -19,6 +19,8 @@ export const registerRadial = (radial: { id: string; items: Omit<RadialItem, 'id
 
 export const getCurrentRadialId = () => exports.ox_lib.getCurrentRadialId();
 
+export const showRadialMenu = (menuId: string) => exports.ox_lib.showRadialMenu(menuId);
+
 export const hideRadial = () => exports.ox_lib.hideRadial();
 
 export const disableRadial = (state: boolean) => exports.ox_lib.disableRadial(state);
--- a/resource/interface/client/radial.lua
+++ b/resource/interface/client/radial.lua
@@ -38,6 +38,10 @@ local menuHistory = {}
 ---@type RadialMenuProps?
 local currentRadial = nil
 
+---The registered menu opened directly with lib.showRadialMenu.
+---@type RadialMenuProps?
+local rootRadial = nil
+
 ---Open a the global radial menu or a registered radial submenu with the given id.
 ---@param id string?
 ---@param option number?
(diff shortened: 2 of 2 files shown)


XCal-Dev:
Related Docs change
https://github.com/overextended/overextended.github.io/pull/282

thelindat:
Please try to reduce code duplication.

XCal-Dev:
Sorry for the delay, I hope this is good! Tested again with both global and directly opened radial menus, all worked fine.

PR #819, opened 2026-07-21
fix: check file existence before LoadResourceFile module probes

qalle-git:
## Problem

Newer FiveM server builds with the sandboxed filesystem log an error for every `LoadResourceFile` call on a file that does not exist. ox_lib's module loaders blind-probe paths that frequently do not exist, so every server start (and many `require` calls) now spam the console:

```
[extra-natives-fi] Failed to open file 'cfx_resource_ox_lib:/imports/require/server.lua' (original path imports/require/server.lua) in resource 'ox_lib'.
[extra-natives-fi] Failed to open file 'cfx_resource_ox_lib:/imports/class/server.lua' (original path imports/class/server.lua) in resource 'ox_lib'.
[extra-natives-fi] Failed to open file 'cfx_resource_ox_lib:/imports/print/server.lua' (original path imports/print/server.lua) in resource 'ox_lib'.
[extra-natives-fi] Failed to open file 'cfx_resource_ox_lib:/imports/array/server.lua' (original path imports/array/server.lua) in resource 'ox_lib'.
```

Three loaders probe blindly:

- `init.lua` `loadModule` tries `imports/<module>/<context>.lua` before falling back to `shared.lua`; many modules only ship `shared.lua`
- `resource/init.lua` lazy `lib.<module>` metatable has the same probe for ox_lib's own bootstrap
- `imports/require/shared.lua` `package.searchpath` tries `?.lua` before `?/init.lua` for every required module

Everything still works (the fallback loads), but the log noise makes real errors easy to miss.

## Fix

List each directory once with the sandbox's `io.readdir` API (already used by `lib.getFilesInDirectory`) and only call `LoadResourceFile` for files that actually exist. Listings are cached per directory.

The guard fails open in every uncertain case, so behaviour is unchanged wherever `io.readdir` cannot answer:

- `io.readdir` missing (older builds / client) falls back to the old probing
- `io.readdir` erroring or returning no handle (e.g. blocked cross-resource reads in `require`) falls back to the old probing

Tested on a live server: the four startup lines above are gone, module loading and `require` (including `@resource.module` and `?/init.lua` layouts) behave as before.

### DIFF
--- a/imports/require/shared.lua
+++ b/imports/require/shared.lua
@@ -50,6 +50,45 @@ end
 
 local tempData = {}
 
+-- Sandboxed builds log an error for every LoadResourceFile miss, so list
+-- directories via io.readdir before probing candidate paths. A failed or
+-- unavailable readdir falls back to probing (listing == false).
+local dirListings = {}
+
+local function resourceFileExists(resource, fileName)
+    if type(io) ~= 'table' or not io.readdir then return true end
+
+    local dir, file = fileName:match('^(.*)/([^/]+)$')
+
+    if not file then
+        dir, file = '', fileName
+    end
+
+    local key = ('@%s/%s'):format(resource, dir)
+    local listing = dirListings[key]
+
+    if listing == nil then
+        listing = false
+        local ok, handle = pcall(io.readdir, key)
+
+        if ok and handle then
+            listing = {}
+
+            for entry in handle:lines() do
+                listing[entry] = true
+            end
+
+            pcall(handle.close, handle)
+        end
+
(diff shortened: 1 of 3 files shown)


unitysync:
no need to make changes to the script when the issue only appears on an early access build with numerous breaking changes.

thelindat:
iolib is not and has never been available for the client on FiveM Legacy, and is not available on Enhanced either. There _isn't_ a way to check if a file exists before attempting to load it. "Solving" the noise on the server is all well and good, but cfx need to sort out their own nonsense before anything changes here.

